PR #2, opened 2019-04-09
Task/API-1074 Add R4 Datatypes and Resources into Repo

esubocz-va:
Create a module for R4 resources and datatypes, add java models, validation, and tests to the module.
Related story: https://vasdvp.atlassian.net/browse/API-1074


r4/src/main/java/gov/va/api/health/r4/api/resources/Coverage.java
```diff
@@ -0,0 +1,241 @@
...
+@Builder
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+@JsonAutoDetect(
+  fieldVisibility = JsonAutoDetect.Visibility.ANY,
+  isGetterVisibility = JsonAutoDetect.Visibility.NONE
+)
+@Schema(
+  description = "https://www.hl7.org/fhir/R4/coverage.html",
+  example = "SWAGGER_EXAMPLE_COVERAGE"
```
dtournour-va: Explain how this works.  Is this just a placeholder because we didn't finish Coverage?

r4/src/main/java/gov/va/api/health/r4/api/validation/AbstractSetFieldCounter.java
```diff
@@ -0,0 +1,58 @@
+package gov.va.api.health.r4.api.validation;
```
dtournour-va: I have a question about this class.

Also as far as Im getting tonight.  Will finish reviewing tomorrow

r4/src/main/java/gov/va/api/health/r4/api/resources/Coverage.java
jhulbert-va: She and I discussed removing this, but we agreed we should wait for the final say by the terminology team and Jason's buy in so we don't delete something to then have to rewrite it.

dtournour-va: I more meant the example="SWAGGER_EXAMPLE_COVERAGE"

r4/src/main/java/gov/va/api/health/r4/api/bundle/BundleLink.java
```diff
@@ -0,0 +1,148 @@
...
+
+  @Valid List<Extension> modifierExtension;
+
+  @NotNull LinkRelation relation;
+
+  @NotBlank
+  @Pattern(regexp = Fhir.URI)
+  String url;
+
+  @SuppressWarnings("unused")
```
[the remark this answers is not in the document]
dtournour-va: This is the only thing I would add

PR #3, opened 2019-04-10
Task/api 1095 extract validation

esubocz-va:
Validation annotations will always be the same for each of the FHIR versions, and hence can all be references from one source. Pull the validation logic into a separate module so it can be accessed by all FHIR version modules.
Related story: https://vasdvp.atlassian.net/browse/API-1095


validation/pom.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>api-starter</artifactId>
```
dtournour-va: I don't think this is right.  What does it need from api-starter?

dtournour-va: I don't think anything in this repo needs to know about anything health-apis related

pom.xml
```diff
@@ -7,9 +7,10 @@
     <version>2.0.0</version>
   </parent>
   <artifactId>fhir-resources-parent</artifactId>
-  <version>1.0.0-SNAPSHOT</version>
+  <version>1.0.1-SNAPSHOT</version>
```
dtournour-va: This is the wrong line.  But I also don't think health-apis-parent is the correct parent for this module either.

PR #17, opened 2019-10-07
API-1561 - Model Claim and CoverageEligibilityRequest

beaugrantham-va:
Added models for `Claim` and `CoverageEligibilityRequest`, following the existing models as templates.

https://www.hl7.org/fhir/claim.html
https://www.hl7.org/fhir/coverageeligibilityrequest.html

This makes no attempt to define an API, it only makes the models available for future use.

Please identify any downstream resources beyond my knowledge that may be affected by these changes.

### DIFF
--- a/r4/src/main/java/gov/va/api/health/r4/api/resources/Claim.java
+++ b/r4/src/main/java/gov/va/api/health/r4/api/resources/Claim.java
@@ -1,0 +1,566 @@
+package gov.va.api.health.r4.api.resources;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import gov.va.api.health.r4.api.Fhir;
+import gov.va.api.health.r4.api.datatypes.Address;
+import gov.va.api.health.r4.api.datatypes.Attachment;
+import gov.va.api.health.r4.api.datatypes.CodeableConcept;
+import gov.va.api.health.r4.api.datatypes.Identifier;
+import gov.va.api.health.r4.api.datatypes.Money;
+import gov.va.api.health.r4.api.datatypes.Period;
+import gov.va.api.health.r4.api.datatypes.Quantity;
+import gov.va.api.health.r4.api.datatypes.SimpleQuantity;
+import gov.va.api.health.r4.api.datatypes.SimpleResource;
+import gov.va.api.health.r4.api.elements.BackboneElement;
+import gov.va.api.health.r4.api.elements.Extension;
+import gov.va.api.health.r4.api.elements.Meta;
+import gov.va.api.health.r4.api.elements.Narrative;
+import gov.va.api.health.r4.api.elements.Reference;
+import gov.va.api.health.validation.api.ExactlyOneOf;
+import gov.va.api.health.validation.api.ZeroOrOneOf;
+import gov.va.api.health.validation.api.ZeroOrOneOfs;
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.util.List;
+import javax.validation.Valid;
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+@JsonAutoDetect(
+  fieldVisibility = JsonAutoDetect.Visibility.ANY,
+  isGetterVisibility = JsonAutoDetect.Visibility.NONE
+)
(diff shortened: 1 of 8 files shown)


beaugrantham-va:
@bschofield-va @jblefkove-va please review.

r4/src/main/java/gov/va/api/health/r4/api/resources/Claim.java
```diff
@@ -0,0 +1,566 @@
...
+
+    @Valid List<Extension> extension;
+
+    @Valid List<Extension> modifierExtension;
+
+    @NotBlank
+    @Pattern(regexp = Fhir.POSITIVE_INT)
+    String sequence;
+
+    @Valid CodeableConcept type;
```
esubocz-va: This should be a `List<CodeableConcept>`

bschofield-va:
Since these are just models, there is no impact to downstream systems. However, if additions were made to an _*Api_ interface, e.g. _CoverageEligibilityResponseApi_, then whatever applications using that would need to be updated to ensure the generated swagger was correct. There is a sample magic that happens with the `SWAGGER_EXAMPLE_*` example values where we programmatically generate sample values in the resulting OpenAPI json.

```
@Schema(description = "https://www.hl7.org/fhir/R4/claim.html", example = "SWAGGER_EXAMPLE_CLAIM")
```

Mechanically, everything looks great, but I think members of your team should verify the correctness of the models, e.g. field names, cardinality, and validation rules.

r4/src/main/java/gov/va/api/health/r4/api/resources/Claim.java
```diff
@@ -0,0 +1,566 @@
...
+    String focal;
+
+    @Valid Identifier identifier;
+
+    @NotNull @Valid Reference coverage;
+
+    @Pattern(regexp = Fhir.STRING)
+    String businessArrangement;
+
+    @Valid List<@Pattern(regexp = Fhir.STRING) String> preAuthRef;
```
esubocz-va: Referencing other resources, I'm seeing Lists of regexp Strings looking like so:
`@Pattern(regexp = Fhir.STRING)
  List<String> preAuthRef;`

r4/src/main/java/gov/va/api/health/r4/api/resources/Claim.java
```diff
@@ -0,0 +1,566 @@
...
+    @Pattern(regexp = Fhir.POSITIVE_INT)
+    String sequence;
+
+    @Valid List<@Pattern(regexp = Fhir.POSITIVE_INT) String> careTeamSequence;
+
+    @Valid List<@Pattern(regexp = Fhir.POSITIVE_INT) String> diagnosisSequence;
+
+    @Valid List<@Pattern(regexp = Fhir.POSITIVE_INT) String> procedureSequence;
+
+    @Valid List<@Pattern(regexp = Fhir.POSITIVE_INT) String> informationSequence;
```
esubocz-va: Same as above.

r4/src/main/java/gov/va/api/health/r4/api/resources/CoverageEligibilityRequest.java
```diff
@@ -0,0 +1,230 @@
...
+  public static class Item implements BackboneElement {
+
+    @Pattern(regexp = Fhir.ID)
+    String id;
+
+    @Valid List<Extension> extension;
+
+    @Valid List<Extension> modifierExtension;
+
+    @Valid List<@Pattern(regexp = Fhir.POSITIVE_INT) String> supportingInfoSequence;
```
esubocz-va: Same as above.

r4/src/main/java/gov/va/api/health/r4/api/resources/Claim.java
bschofield-va: The original looks correct to me. `List<@Pattern(regexp = Fhir.STRING) String>`
https://beanvalidation.org/2.0/spec/#constraintdeclarationvalidationprocess-containerelementconstraints

esubocz-va: Think we may have some issues in other resources then, as I got that example from `r4/Capability`

r4/src/main/java/gov/va/api/health/r4/api/resources/CoverageEligibilityRequest.java
```diff
@@ -0,0 +1,230 @@
...
+  public static class SupportingInfo implements BackboneElement {
+
+    @Pattern(regexp = Fhir.ID)
+    String id;
+
+    @Valid List<Extension> extension;
+
+    @Valid List<Extension> modifierExtension;
+
+    @NotBlank
```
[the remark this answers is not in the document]
beaugrantham-va: If the question is about whether or not it's required, the spec says yes.
https://www.hl7.org/fhir/R4/coverageeligibilityrequest-definitions.html#CoverageEligibilityRequest.supportingInfo.sequence

If the question is about whether or not this is the appropriate annotation, it follows the pattern of other resources that have similar required fields.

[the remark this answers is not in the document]
dtournour-va: It was the second one.  If whether it's the appropriate annotation.  Ok . We can resolve this.

PR #23, opened 2019-10-10
Add model for ServiceRequest

beaugrantham-va:


r4/src/main/java/gov/va/api/health/r4/api/resources/ServiceRequest.java
```diff
@@ -0,0 +1,188 @@
...
+    @JsonProperty("entered-in-error")
+    entered_in_error,
+    cancelled;
+  }
+
+  public enum Intent {
+    proposal,
+    plan,
+    order;
+  }
```
esubocz-va: there are a few more codes if you follow the `RequestIntent` link next to intent on https://www.hl7.org/fhir/servicerequest.html

esubocz-va:
The same datatypes are added in the MedicationRequest PR that I've reviewed. We can just merge that PR first & merge master in this branch & it'll be ready to go after the little bit of feedback is addressed.

r4/src/main/java/gov/va/api/health/r4/api/resources/ServiceRequest.java
beaugrantham-va: I struggled with this; the spec was confusing.

The description on the [resource content](https://www.hl7.org/fhir/servicerequest.html#resource) says

> proposal | plan | order +

The [detailed description](https://www.hl7.org/fhir/servicerequest-definitions.html#ServiceRequest.intent) says

> Whether the request is a proposal, plan, an original order or a reflex order.

And the [value set](https://www.hl7.org/fhir/valueset-request-intent.html) lists even more additional values.

I guess I can expand it to the full list and it can be narrowed down once these models are actually used in the future.

PR #24, opened 2019-10-11
add ExplanationOfBenefit resource model and tests

esubocz-va:
Add R4 model and tests for ExplanationOfBenefit resource.
Resource: https://www.hl7.org/fhir/explanationofbenefit.html
Related task: https://vasdvp.atlassian.net/browse/API-1573

### DIFF
--- a/r4/src/main/java/gov/va/api/health/r4/api/resources/ExplanationOfBenefit.java
+++ b/r4/src/main/java/gov/va/api/health/r4/api/resources/ExplanationOfBenefit.java
@@ -42,7 +42,7 @@ import lombok.NoArgsConstructor;
   fieldVisibility = JsonAutoDetect.Visibility.ANY,
   isGetterVisibility = JsonAutoDetect.Visibility.NONE
 )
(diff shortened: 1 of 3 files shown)


r4/src/main/java/gov/va/api/health/r4/api/resources/ExplanationOfBenefit.java
```diff
@@ -0,0 +1,932 @@
...
+
+@Data
+@Builder
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+@JsonAutoDetect(
+  fieldVisibility = JsonAutoDetect.Visibility.ANY,
+  isGetterVisibility = JsonAutoDetect.Visibility.NONE
+)
+@Schema(description = "https://www.hl7.org/fhir/R4/claim.html", example = "SWAGGER_EXAMPLE_CLAIM")
```
beaugrantham-va: Schema description and example need to be updated.

PR #30, opened 2019-11-06
Update Claim with Entry and Bundle and AbstractBundle total.

lbrown-va:
Update Claim with Entry and Bundle.  Also, found that the r4/AbstractBundle, total should be an Integer.  This change was made for consistency with other resources and the fhir spec.

### DIFF
--- a/r4/src/main/java/gov/va/api/health/r4/api/bundle/AbstractBundle.java
+++ b/r4/src/main/java/gov/va/api/health/r4/api/bundle/AbstractBundle.java
@@ -10,6 +10,7 @@ import gov.va.api.health.r4.api.resources.Resource;
 import io.swagger.v3.oas.annotations.media.Schema;
 import java.util.List;
 import javax.validation.Valid;
+import javax.validation.constraints.Min;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Pattern;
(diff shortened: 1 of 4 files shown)


r4/src/main/java/gov/va/api/health/r4/api/resources/Claim.java
```diff
@@ -138,6 +145,68 @@
...
+    }
+  }
+
+  @Data
+  @NoArgsConstructor
+  @EqualsAndHashCode(callSuper = true)
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  @JsonDeserialize(builder = Claim.Bundle.BundleBuilder.class)
+  @Schema(name = "ClaimBundle", example = "SWAGGER_EXAMPLE_CLAIM_BUNDLE")
+  public static class Bundle extends AbstractBundle<CoverageEligibilityResponse.Entry> {
```
jblefkove-va: copy-paste error?

lbrown-va: Oopsy.

r4/src/main/java/gov/va/api/health/r4/api/bundle/AbstractBundle.java
```diff
@@ -43,8 +44,8 @@
   @Pattern(regexp = Fhir.INSTANT)
   protected String timestamp;
 
-  @Pattern(regexp = Fhir.UNSIGNED_INT)
-  protected String total;
+  @Min(0)
+  protected Integer total;
```
bschofield-va: This is the right move ... but it will break r4 consumers.

PR #33, opened 2019-11-13
Rename to eliminate conflicts when same service implements multiple api.

lbrown-va:


r4/src/main/java/gov/va/api/health/r4/api/CoverageApi.java
```diff
@@ -84,7 +84,7 @@
           schema = @Schema(implementation = OperationOutcome.class)
         )
   )
-  Coverage.Bundle searchByPatient(
+  Coverage.Bundle coverageSearch(
```
bschofield-va: For what it's worth, `searchByPatient` is more inline with the other resources.  Given this is `CoverageApi`, the _coverage _ is a bit redundant. Is the intention to expand `coverageSearch` to support _all_ 11 search parameters? Or we will eventually see other specialized searches, e.g. `coverageSearchByBeneficiary` or `coverageSearchByDependent`.

bschofield-va: Is there a conflict because you're using one implementation for both interfaces?

lbrown-va: That's what I originally had attempted, but when I started merging branches together this bombs:
```
public interface ClaimService extends ClaimApi, CoverageApi, ExplanationOfBenefitApi, MetadataApi {
```
So I tried following what I thought was the data-query convention as a work around.

PR #35, opened 2019-11-20
PIV-36 - Add r4 examples

beaugrantham-va:
See also 
* https://github.com/department-of-veterans-affairs/health-apis-claims/pull/14
* https://github.com/department-of-veterans-affairs/health-apis-urgent-care-eligibility/pull/58


beaugrantham-va:
> 
> 
>     1. r4-examples/README, usage might be helpful?
> 
>     2. r4-examples/pom.xml, I wonder if parent should be something other than api-starter?
> 
>     3. Thank you

I added an example usage to the README.

I considered a different parent too, but I ended up followed the convention used by all the other modules within this repo. I'll defer to the repository owners.

r4-examples/src/main/java/gov/va/api/health/r4/api/swaggerexamples/SwaggerOperationOutcome.java
```diff
@@ -0,0 +1,30 @@
...
+   * @return an example OperationOutcome.
+   */
+  public static OperationOutcome operationOutcome() {
+    return OperationOutcome.builder()
+        .resourceType("OperationOutcome")
+        .issue(
+            asList(
+                OperationOutcome.Issue.builder()
+                    .severity(OperationOutcome.Issue.IssueSeverity.error)
+                    .code("forbidden")
```
bschofield-va: I wonder if this should be more generic. 
When it ends up on `developer.va.gov`, the OO example gets used for each non-200 response. We end up with an OO example for `400` _Bad Request_ that says _forbidden_.

Thoughts?

bschofield-va:
> 2. r4-examples/pom.xml, I wonder if parent should be something other than api-starter?

Looks like `health-apis-parent` itself would work fine for this module.

r4/src/main/java/gov/va/api/health/r4/api/resources/OperationOutcome.java
```diff
@@ -28,7 +28,8 @@
 @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
```
bschofield-va: This is going to break any other R4 consumer.

r4-examples/src/main/java/gov/va/api/health/r4/api/swaggerexamples/SwaggerOperationOutcome.java
beaugrantham-va: This example is already published. This story just moves it out of urgent-care into this common project so others can use it.

https://api.va.gov/services/fhir/v0/r4/openapi.json

It's populated under the generic schemas section (`$.components.schemas.OperationOutcome`) and is not tied to a particular request type.

That said, I'm happy to update it to whatever you like.

r4/src/main/java/gov/va/api/health/r4/api/resources/OperationOutcome.java
beaugrantham-va: See related PRs in the description. UC and Claims have PRs to use this new library.

r4-examples/src/main/java/gov/va/api/health/r4/api/swaggerexamples/SwaggerOperationOutcome.java
bschofield-va: How about ...

```
"code" : "request error"
...
   "text" : "This request can not be processed"
```

PR #49, opened 2020-05-04
Shank 595 condition model

jhulbert-va:
https://vasdvp.atlassian.net/browse/SHANK-595


us-core-r4/src/main/java/gov/va/api/health/uscorer4/api/ConditionApi.java
```diff
@@ -0,0 +1,105 @@
...
+  @Operation(
+      summary = "Condition Search",
+      description =
+          "https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-condition.html",
+      tags = {"Condition"})
+  @GET
+  @Path("Condition")
+  @ApiResponse(
+      responseCode = "200",
+      description = "Record Found",
```
jblefkove-va, line 63: plural?

PR #50, opened 2020-05-05
pull out argonaut's copy of ARFV to validation module

eclendenning-va:


validation/pom.xml
```diff
@@ -15,5 +15,15 @@
...
     </dependency>
+    <dependency>
+      <groupId>gov.va.api.health</groupId>
+      <artifactId>service-auto-config</artifactId>
+      <version>5.0.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
```
jblefkove-va, line 26: `compile` scope for this dependency is suspicious

validation/src/test/java/gov/va/api/health/validation/api/ZeroOrOneOfVerifierTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.opentest4j.AssertionFailedError;
+
+public class ZeroOrOneOfVerifierTest {
+  @Test(expected = AssertionFailedError.class)
+  public void allFields() {
+    Map<String, Supplier<?>> stringTypes = new HashMap<>();
+    Map<Class<?>, Supplier<?>> types = new HashMap<>();
+    types.put(String.class, () -> "hello");
+    types.put(Integer.class, () -> 1);
+    stringTypes.put("", () -> "hello");
```
jblefkove-va, line 18: Use `ImmutableMap.of("", () -> "hello")` to create simple maps in one line. (Will probably need to add a dependency on Guava to this project.)

jblefkove-va:
There is `ExactlyOneOfExtensionVerifier` in argonaut and r4... do we want to consolidate that as well?

jblefkove-va:
398 lines deleted 👍

validation/pom.xml
```diff
@@ -15,5 +15,20 @@
       <groupId>org.springframework</groupId>
       <artifactId>spring-beans</artifactId>
     </dependency>
+    <dependency>
+      <groupId>gov.va.api.health</groupId>
+      <artifactId>service-auto-config</artifactId>
+      <version>5.0.0-SNAPSHOT</version>
```
jblefkove-va, line 21: Move this into a variable

validation/pom.xml
```diff
@@ -15,5 +15,20 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
```
jblefkove-va, line 25: Still not crazy about this dependency. It looks like it's only used for one assertion. Let's replace that assertion with a `Preconditions.checkState` from Guava.

validation/pom.xml
jblefkove-va: Actually it looks like this is only used in one place, to get `JacksonConfig` to create an object mapper for serializing something into a log statement. I think you can omit this and just do `new ObjectMapper()` instead.

PR #51, opened 2020-05-11
API 822 unify common

bschofield-va:
To support modules that can process dstu2, stu3, and r4 references the same, this introduces
- an  IsReference interface
- a new fhir-common module with standardized date time parsing

### DIFF
--- a/fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirDateTime.java
+++ b/fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirDateTime.java
@@ -1,0 +1,26 @@
+package gov.va.api.health.fhir.api;
+
+import java.time.Instant;
+import javax.xml.datatype.DatatypeFactory;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import lombok.SneakyThrows;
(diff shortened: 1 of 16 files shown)


fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirDateTime.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.time.Instant;
+import javax.xml.datatype.DatatypeFactory;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import lombok.SneakyThrows;
+import org.apache.commons.lang3.StringUtils;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class FhirDateTime {
```
taylorrassmannlibertyits, line 11: Private definition and right below is public?

bschofield-va: the no args constructor is private to prevent anyone from instantiating the utility class

taylorrassmannlibertyits: Thanks!

PR #53, opened 2020-05-11
API 279 r4 fix capability

bschofield-va:
- 3.0.0 Breaking changes to R4 Capability - Renames to `CapabilityStatement` - Fixes `@Pattern` constraints incorrectly applied to `List<String>`
- Fix Capability List of Strings with @Pattern constraints
- Fixes `String` type declarations for Boolean, Integers, and Decimal type fields to be correct data types.

`CapabilityStatement` is the only file with real changes. That is, all of the files are changed due to renaming.


bschofield-va:
This PR contains breaking changes that will affect the Unifier. Structurally, R4 CapabilityStatement is the same, but it has been renamed to reflect it's R4 resource type.

@beaugrantham-va @dbrown-va @esubocz-va

r4/src/main/java/gov/va/api/health/r4/api/MetadataApi.java
```diff
@@ -23,7 +23,7 @@
       content =
           @Content(
               mediaType = "application/json+fhir",
-              schema = @Schema(implementation = Capability.class),
+              schema = @Schema(implementation = CapabilityStatement.class),
```
jblefkove-va, line 26: Should the summary on line 15 also be CapabilityStatement?

bschofield-va: OpenAPI declares ..."

> operationId is an optional unique string used to identify an operation. If provided, these IDs must be unique among all operations described in your API.

There isn't really a _wrong_ answer, but I think we should change it. 
The example uses `metadata`, https://www.hl7.org/fhir/us/core/us-core-server.openapi.json

PR #59, opened 2020-05-14
patient is a uscorer4 resource not r4

eclendenning-va:


us-core-r4/src/main/java/gov/va/api/health/uscorer4/api/resources/Patient.java
```diff
@@ -52,7 +52,7 @@
 @Schema(
     description =
         "https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-patient.html",
-    example = "${r4.patient:com.example.Example#example}")
+    example = "${uscorer4.patient:com.example.Example#example}")
```
jblefkove-va, line 55: com.example?

us-core-r4/src/main/java/gov/va/api/health/uscorer4/api/resources/Patient.java
```diff
@@ -184,7 +184,7 @@ private boolean isValidUsCoreExtensionCount(String url, int maxAllowedOmbExtensi
   @Schema(
       name = "PatientBundle",
       example =
-          "${r4.patientBundle:gov.va.api.health.r4.api.swaggerexamples."
+          "${uscorer4.patientBundle:gov.va.api.health.r4.api.swaggerexamples."
```
jblefkove-va, line 187: Should also be `uscorer4` in the package

PR #67, opened 2020-06-02
API-1037-r4-observation-openapi

jhulbert-va:
[API-1037](https://vajira.max.gov/browse/API-1037)

[R4 Observation Controller](https://github.com/department-of-veterans-affairs/health-apis-data-query/blob/master/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/R4ObservationController.java)


taylorrassmannlibertyits:
Do any samples or swagger code need to change for this? Or expand on what was added?

jhulbert-va:
> Do any samples or swagger code need to change for this? Or expand on what was added?

No. Those new params will still return the same result as in the example, it will just return a smaller subset.

PR #73, opened 2020-07-20
Api 1572 diagnostic report r4 model

jhulbert-va:
[API-1572](https://vajira.max.gov/browse/API-1572)

### DIFF
--- a/us-core-r4/src/main/java/gov/va/api/health/uscorer4/api/resources/DiagnosticReport.java
+++ b/us-core-r4/src/main/java/gov/va/api/health/uscorer4/api/resources/DiagnosticReport.java
@@ -1,0 +1,247 @@
+package gov.va.api.health.uscorer4.api.resources;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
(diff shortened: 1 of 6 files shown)


us-core-r4/src/main/java/gov/va/api/health/uscorer4/api/resources/DiagnosticReport.java
```diff
@@ -0,0 +1,252 @@
...
+  @Valid List<Identifier> identifier;
+
+  @Valid List<Reference> basedOn;
+
+  @NotNull DiagnosticReportStatus status;
+
+  /* This is a slice definition.
+   * That being the case, it requires fields that plain old CC would not.
+   */
+  @Valid @NotEmpty List<CodeableConcept> category;
```
ilaflamme-va, line 94: I don't think this is good enough.
A slice definition states that it _must_ contain a `coding`, which in turn _must_ contain a `system` and `code`. Our validation misses that, here.

us-core-r4/src/main/java/gov/va/api/health/uscorer4/api/resources/DiagnosticReport.java
```diff
@@ -0,0 +1,252 @@
...
+
+  @Valid Period effectivePeriod;
+
+  @NotNull
+  @Pattern(regexp = Fhir.INSTANT)
+  String issued;
+
+  @Valid List<Reference> performer;
+
+  @Valid Extension _performer;
```
ilaflamme-va, line 113: We don't need a DAR here, as it's an optional field

PR #75, opened 2020-07-23
Api 1627 dates as instant

jhulbert-va:
https://vajira.max.gov/browse/API-1627


bschofield-va:
https://www.hl7.org/fhir/datatypes.html#dateTime

A date, date-time or partial date (e.g. just year or year + month) as used in human communication. The format is YYYY, YYYY-MM, YYYY-MM-DD or YYYY-MM-DDThh:mm:ss+zz:zz, e.g. 2018, 1973-06, 1905-08-23, 2015-02-07T13:28:17-05:00 or 2017-01-01T00:00:00.000Z. If hours and minutes are specified, a time zone SHALL be populated. Seconds must be provided due to schema type constraints but may be zero-filled and may be ignored at receiver discretion. Dates SHALL be valid dates. The time "24:00" is not allowed. Leap Seconds are allowed 

```
jshell> Instant.parse("2005")
|  Exception java.time.format.DateTimeParseException: Text '2005' could not be parsed at index 4
```

PR #76, opened 2020-08-11
Api 1713 uscorer4 organization model

ilaflamme-va:


us-core-r4/src/test/java/gov/va/api/health/uscorer4/api/resources/OrganizationTest.java
```diff
@@ -0,0 +1,76 @@
...
+    assertRoundTrip(bundle);
+  }
+
+  @Test
+  public void organization() {
+    assertRoundTrip(data.organization());
+  }
+
+  @Test
+  public void validationFailsGivenBadIdentifierSlice() {
```
jhulbert-va, line 61: I think there should be a check for when the identifier is null as well so that we can check all cases. `data.organization()` `roundTrip()` will check when only one exists.

us-core-r4/src/main/java/gov/va/api/health/uscorer4/api/resources/Organization.java
```diff
@@ -0,0 +1,212 @@
...
+  */
+  @Valid List<Identifier> identifier;
+
+  @NotNull Boolean active;
+
+  @Valid List<CodeableConcept> type;
+
+  @Valid @NotNull String name;
+
+  @Valid String alias;
```
jhulbert-va, line 88: Should be a list.

us-core-r4/src/main/java/gov/va/api/health/uscorer4/api/resources/Organization.java
```diff
@@ -0,0 +1,212 @@
...
+  but at most there can be one Identifier NPI slice,
+  and at most there can be one Identifier CLIA slice.
+  */
+  @Valid List<Identifier> identifier;
+
+  @NotNull Boolean active;
+
+  @Valid List<CodeableConcept> type;
+
+  @Valid @NotNull String name;
```
jhulbert-va, line 86: Strings dont need the `@Valid` annotation.

us-core-r4/src/main/java/gov/va/api/health/uscorer4/api/resources/Organization.java
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  @Data
+  @Builder
+  @Schema(name = "OrganizationContact")
+  @AllArgsConstructor
+  @NoArgsConstructor(access = AccessLevel.PRIVATE)
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  public static class Contact implements BackboneElement {
+    String id;
```
jhulbert-va, line 133: `@Pattern()`

PR #89, opened 2020-09-11
Moved a field that was incorrectly placed in R4 Coverage Class

that-guy-is-it:


jblefkove-va:
`@NotEmpty List<Reference> payor;` needs `@valid`

jblefkove-va (changes requested):
inline

jblefkove-va:
`coverageClass` also missing `@Valid`

jblefkove-va:
`@NotNull CodeableConcept type;` in Exception missing `@Valid`

jblefkove-va:
` @NotNull CodeableConcept type;` in CoverageClass missing `@Valid`

jblefkove-va:
Should `Exception` be an inner class of `CostToBeneficiary`?

r4/src/main/java/gov/va/api/health/r4/api/resources/Coverage.java
```diff
@@ -117,7 +119,8 @@
   @Schema(
       name = "CoverageBundle",
       example =
-          "${r4.coverageBundle:gov.va.api.health.r4.api.swaggerexamples.SwaggerCoverage#coverageBundle}")
+          "${r4.coverageBundle:gov.va.api.health.r4.api."
+              + "swaggerexamples.SwaggerCoverage#coverageBundle}")
```
jblefkove-va, line 123: Take a look at `checkstyle-suppressions.xml`. There is a suppression for this long line, hardcoded to line 120. You can remove that now.

jblefkove-va (approved):

jblefkove-va (approved):

PR #93, opened 2020-09-24
Add CarinBlueButton annotation

bschofield-va:


jhulbert-va (commented):
Should the package be different?

`gov.va.api.health.r4.api.specifications`

bschofield-va:
> Should the package be different?
> 
> `gov.va.api.health.r4.api.specifications`

I thought about putting in a `.ig` package, but we have `FHIR` sitting at that same level. I could be convinced to move it, but it didn't seem to bring too much value. Curious what others think.

ilaflamme-va:
I'm ok with it, as it.

jhulbert-va (approved):

eclendenning-va (approved):

PR #95, opened 2020-09-24
uscore annotations added to organization model

ilaflamme-va:
useful links:
https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-organization.html
http://hl7.org/fhir/R4/organization.html

### DIFF
--- a/us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
+++ b/us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import gov.va.api.health.r4.api.Fhir;
+import gov.va.api.health.r4.api.UsCore;
 import gov.va.api.health.r4.api.bundle.AbstractBundle;
 import gov.va.api.health.r4.api.bundle.AbstractEntry;
 import gov.va.api.health.r4.api.bundle.BundleLink;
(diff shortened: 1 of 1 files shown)


us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
```diff
@@ -96,6 +100,7 @@
 
   @Valid List<Reference> endpoint;
 
+  @UsCore
```
jblefkove-va, line 103: `identifier:CLIA` is not marked as `must be supported`

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
```diff
@@ -78,16 +79,19 @@
...
   @NotNull Boolean active;
 
   @Valid List<CodeableConcept> type;
 
+  @UsCore
   @NotNull String name;
 
   @Valid List<String> alias;
 
   @Valid List<ContactPoint> telecom;
```
jblefkove-va, line 92: telecom must be supported?

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
ilaflamme-va: We aren't annotating fields that must be supported (because null is valid). The annotations are for places the models have changed to support us-core, as an upgrade on top of r4.

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
ilaflamme-va: There is no change to the model for this, because a null value is legal, even though it says "Must Be Supported".
Must Be Supported does not translate to required.

ilaflamme-va: The model remains the same for both R4 and UsCoreR4 in this case.

jblefkove-va (approved):

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
```diff
@@ -78,16 +79,19 @@
   */
   @Valid List<Identifier> identifier;
 
+  @UsCore
```
jhulbert-va, line 82: Update the annotations with new fields.

jhulbert-va (changes requested):

ilaflamme-va:
Closing this, as the focus is on CarinBB at this time.

PR #97, opened 2020-09-25
add cbb organization annotations

ilaflamme-va:


us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
```diff
@@ -76,7 +77,14 @@
   but at most there can be one Identifier NPI slice,
   and at most there can be one Identifier CLIA slice.
   */
-  @Valid List<Identifier> identifier;
+  @CarinBlueButton(
+      note =
+          "Carin includes an additional slice definition for Identifier:TIN. "
+              + "The Identifier.type field's cardinality changes cardinality to 1..1. "
+              + "The Identifier:NPI slice's Type.Coding changes cardinality to 1..*. "
+              + "The Identifier:NPI slice's Type.Coding.Code changes cardinality to 1..1.")
```
jhulbert-va, line 85: `identifier:TIN` and `identifier:NPI` are each their own slice. NPI is `0..1` and TIN is `0..*` The number of allowed slices for each should be listed here as well as the internal changes to the identifier definition.

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
```diff
@@ -76,7 +77,14 @@
   but at most there can be one Identifier NPI slice,
   and at most there can be one Identifier CLIA slice.
   */
-  @Valid List<Identifier> identifier;
+  @CarinBlueButton(
+      note =
+          "Carin includes an additional slice definition for Identifier:TIN with cardinality 0..*."
+              + "The Identifier.type field's cardinality changes cardinality to 1..1. "
+              + "The Identifier:NPI slice's Type.Coding changes cardinality to 1..*. "
+              + "The Identifier:NPI slice's Type.Coding.Code changes cardinality to 1..1.")
```
jhulbert-va, line 85: This is the case with the TIN slice as well.

jhulbert-va (approved):

PR #99, opened 2020-10-02
Api 2639 organization address line

jjabando1:

PR #101, opened 2020-10-21
Added Practitioner class

that-guy-is-it:


us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,99 @@
...
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+@JsonAutoDetect(
+    fieldVisibility = JsonAutoDetect.Visibility.ANY,
+    isGetterVisibility = JsonAutoDetect.Visibility.NONE)
+@Schema(description = "https://www.hl7.org/fhir/practitioner.html")
```
jblefkove-va, line 34: US Core R4 practitioner: http://hl7.org/fhir/us/core/StructureDefinition-us-core-practitioner.html

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,99 @@
...
+  @Pattern(regexp = Fhir.ID)
+  String id;
+
+  @Valid Meta meta;
+
+  @Pattern(regexp = Fhir.URI)
+  String implicitRules;
+
+  @Pattern(regexp = Fhir.CODE)
+  String language;
```
jblefkove-va, line 46: missing `text`, `contained`, `extension`, `modifierExtension`

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,99 @@
...
+
+  @Valid List<Identifier> identifier;
```
jblefkove-va, line 48: not empty

jblefkove-va, line 48: `system` and `value` are required fields inside this `Identifier`, although vanilla R4 does not require them. @ilaflamme-va I think you ran into a similar problem recently? Should this class have its own `Identifier` class with more strict validation?

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,99 @@
...
+
+  Boolean active;
+
+  @Valid List<HumanName> name;
```
jblefkove-va, line 52: not empty

jblefkove-va: similar to above, `family` is a required field for `HumanName` on this class, but not for general R4

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,99 @@
...
+  String date;
+
+  @Valid List<Attachment> photo;
+
+  @Valid List<Qualification> qualification;
+
+  @Valid List<CodeableConcept> communication;
+
+  enum GenderCode {
+    MALE,
```
jblefkove-va, line 70: lowercase

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,99 @@
...
+    @Pattern(regexp = Fhir.ID)
+    String id;
+
+    @Valid List<Extension> extension;
+
+    @Valid List<Extension> modifierExtension;
+
+    @Valid List<Identifier> identifier;
+
+    @NotNull CodeableConcept code;
```
jblefkove-va, line 93: `@Valid`

jblefkove-va:
Missing `Bundle` and `Entry` classes

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
jblefkove-va: Found it: https://github.com/department-of-veterans-affairs/health-apis-fhir-resources/pull/99/files

So this class can use the shared R4 classes for `Identifier` and `HumanName`. The additional validations can be specified in custom validation methods on this class.

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,197 @@
...
+
+  @Valid List<Attachment> photo;
+
+  @Valid List<Qualification> qualification;
+
+  @Valid List<CodeableConcept> communication;
+
+  @JsonIgnore
+  @SuppressWarnings("unused")
+  @AssertTrue(message = "At most one IdentifierClia can be specified.")
```
jblefkove-va, line 90: this message is not accurate

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,197 @@
...
+  private boolean isValidIdentifier() {
+    if (identifier == null) {
+      return false;
+    }
+    return identifier.stream().filter(e -> e.system() != null && e.value() != null).count() >= 1;
```
jblefkove-va, line 95: should verify that *every* identifier has system and value set

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  @JsonIgnore
+  @SuppressWarnings("unused")
+  @AssertTrue(message = "At most one IdentifierClia can be specified.")
```
jblefkove-va, line 100: inaccurate message

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,197 @@
...
+  private boolean isValidName() {
+    if (name == null) {
+      return false;
+    }
+    return name.stream().filter(e -> e.family() != null).count() >= 1;
```
jblefkove-va, line 105: similarly this should verify that every name has `family `set

jblefkove-va:
Needs a unit test, plus tests for the custom validations

us-core-r4/src/test/java/gov/va/api/health/r4/api/samples/SamplePractitioners.java
```diff
@@ -0,0 +1,33 @@
...
+import gov.va.api.health.r4.api.resources.Practitioner;
+import java.util.Collections;
+import java.util.List;
+import lombok.NoArgsConstructor;
+import lombok.experimental.Delegate;
+
+@NoArgsConstructor(staticName = "get")
+public class SamplePractitioners {
+
+  @Delegate SampleDataTypes dataTypes = SampleDataTypes.get();
```
jblefkove-va, line 12: Remove this if it is not used

us-core-r4/src/test/java/gov/va/api/health/r4/api/resources/PractitionerTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+  @Test
+  public void validationPassesGivenGoodIdentifier() {
+    Practitioner prac = data.practitioner();
+    assertThat(violationsOf(prac)).isEmpty();
+    prac.identifier().stream()
+        .forEach(
+            i -> {
+              i.value("value");
+              i.system("system");
```
jblefkove-va, line 85: you can chain these together... `i.value(...).system(...)`

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,202 @@
...
+
+  @Valid Narrative text;
+
+  @Valid List<SimpleResource> contained;
+
+  @Valid List<Extension> extensions;
+
+  @Valid List<Extension> modifierExtensions;
+
+  @Valid @NotEmpty List<Identifier> identifier;
```
ilaflamme-va, line 71: missing validation method for an Identifier:NPI slice.

ilaflamme-va (changes requested):

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
ilaflamme-va: Your current validation method will throw away good Identifier:NPI slices

us-core-r4/src/test/java/gov/va/api/health/r4/api/samples/SamplePractitioners.java
that-guy-is-it: It is used, though. The delegate is how I get access to the constructors of the types, such as `meta()` and `narrative()`. Is there another way?

jblefkove-va: My mistake, I didn't see references to it but missed the `@Delegate`

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,221 @@
...
+  @AssertTrue(message = "At most one IdentifierNpi can be specified.")
+  private boolean isValidIdentifierNpiSlice() {
+    if (identifier == null) {
+      return true;
+    }
+    return identifier.stream()
+            .filter(
+                e ->
+                    StringUtils.isNotBlank(e.system())
+                        && e.system().equals("http://hl7.org/fhir/sid/us-npi"))
```
jblefkove-va, line 103: how about `"http://hl7.org/fhir/sid/us-npi".equals(e.system())` ?

that-guy-is-it: Hmm, I am confused, how would that be better?

jblefkove-va: it eliminates the first check

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,214 @@
...
+                e ->
+                    StringUtils.isNotEmpty(e.value())
+                        && StringUtils.isNotEmpty(e.system())
+                        && !"http://hl7.org/fhir/sid/us-npi".equals(e.system()));
+
+    boolean validNpiSlice =
+        identifier.stream().filter(e -> "http://hl7.org/fhir/sid/us-npi".equals(e.system())).count()
+            == 1;
+
+    return (validAllSlice ^ validNpiSlice);
```
jblefkove-va, line 112: It's less clear to me now what this method is checking? Are these the conditions?

1. At least one identifier is present
2. Every identifier has system and value set, unless system is NPI
3. At most one identifier has system NPI

If so, I think it is simpler to keep the conditions separate and return early if necessary, i.e:

1. Return false if there are no identifiers (method already does this)
2. Return false if there is more than one NPI identifier
3. Return true/false if every identifier has system and value set, filtering out those where system is NPI

bschofield-va: 10/10 agree, this fails Occam's razor.
I value simple to read code. No points for cleverness.

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -0,0 +1,212 @@
...
+        > 1) {
+      return false;
+    }
+
+    return identifier.stream()
+        .allMatch(
+            e ->
+                StringUtils.isNotEmpty(e.system())
+                    && (StringUtils.isNotEmpty(e.value())
+                        || "http://hl7.org/fhir/sid/us-npi".equals(e.system())));
```
jblefkove-va, line 110: You can simplify this check by filtering out the NPI entries first:
```
    return identifier
        .stream()
        .filter(e -> !"http://hl7.org/fhir/sid/us-npi".equals(e.system()))
        .allMatch(e -> isNotEmpty(e.system()) && isNotEmpty(e.value()));
```

(Also, we are not fully consistent about it, but we often statically import from `StringUtils`. In this case that definitely benefits readability.)

jblefkove-va (approved):

ilaflamme-va (approved):

PR #106, opened 2020-11-20
CodeableConcepts for all FHIR versions will now assume a text value of their inner codings if the display is not explicitly set.

bschofield-va:


eclendenning-va (approved):

bschofield-va:
FYI @jblefkove-va

PR #111, opened 2021-01-28
API-4376: immunization vaccine protocol

jjabando1:


argonaut-dstu2/src/test/java/gov/va/api/health/dstu2/api/samples/SampleKnownTypes.java
```diff
@@ -43,6 +44,14 @@
     return suppliers;
   }
 
+  public Map<Class<?>, Supplier<?>> knownTypeCodeableConceptList() {
```
jhulbert-va, line 47: This feels kinda one-off. I think instead of putting it in this reusable class, it should just be defined in `ImmunizationTest`.
Soemthing like:
```
Map.of(
Extension.class, dataTypes::extension,
List.class, dataTypes::codeableConceptList
)
```

eclendenning-va (approved):

jhulbert-va (approved):

PR #112, opened 2021-02-09
API 4721 Mixed FHIR Custom Deserializer

that-guy-is-it:


us-core-r4/src/test/java/gov/va/api/health/r4/api/bundle/MixedBundleTest.java
```diff
@@ -0,0 +1,282 @@
...
+import gov.va.api.health.r4.api.samples.SampleRelatedPersons;
+import gov.va.api.health.r4.api.samples.SampleTerminologyCapabilities;
+import java.util.Arrays;
+import java.util.List;
+import lombok.SneakyThrows;
+import org.junit.jupiter.api.Test;
+
+public class MixedBundleTest {
+  @Test
+  void _mixed() {
```
jblefkove-va, line 66: This test is now redundant, the individual resource round-trip tests have it covered

us-core-r4/src/test/java/gov/va/api/health/r4/api/bundle/MixedBundleTest.java
```diff
@@ -0,0 +1,282 @@
...
+    Observation r = SampleObservations.get().observation();
+    assertRoundTrip(
+        MixedBundle.builder().entry(List.of(MixedEntry.builder().resource(r).build())).build());
+  }
+
+  @Test
+  void operationOutcome() {
+    SampleDataTypes data = SampleDataTypes.get();
+    OperationOutcome r =
+        OperationOutcome.builder()
```
jblefkove-va, line 200: Refactor this into `SampleDataTypes`

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Resource.java
```diff
@@ -17,4 +28,59 @@
...
+
+    @Override
+    @SneakyThrows
+    public Resource deserialize(JsonParser jp, DeserializationContext ctxt) {
+      ObjectMapper mapper = (ObjectMapper) jp.getCodec();
+      ObjectNode root = mapper.readTree(jp);
+      String type = root.get("resourceType").asText();
+      Class<? extends Resource> clazz = CLASS_MAP.get(type);
+      if (clazz != null) {
+        return (Resource) mapper.readValue(root.toString(), clazz);
```
jblefkove-va, line 81: unnecessary cast?

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Resource.java
```diff
@@ -17,4 +28,59 @@
...
 
   Meta meta();
+
+  public final class ResourceDeserializer extends StdDeserializer<Resource> {
+
+    static final Map<String, Class<? extends Resource>> CLASS_MAP =
+        Map.ofEntries(
+            Map.entry("AllergyIntolerance", AllergyIntolerance.class),
+            Map.entry("Appointment", Appointment.class),
+            Map.entry("Bundle", MixedBundle.class),
```
jblefkove-va, line 38: This entry is not necessary. This solution (vs annotations) sidesteps Bundles being deserialized as Resources.

us-core-r4/src/test/java/gov/va/api/health/r4/api/bundle/MixedBundleTest.java
```diff
@@ -0,0 +1,282 @@
...
+
+  @Test
+  void allergyIntolerance() {
+    AllergyIntolerance r = SampleAllergyIntolerances.get().allergyIntolerance();
+    assertRoundTrip(
+        MixedBundle.builder().entry(List.of(MixedEntry.builder().resource(r).build())).build());
+  }
+
+  @Test
+  void appoointment() {
```
jblefkove-va, line 91: typo

us-core-r4/src/test/java/gov/va/api/health/r4/api/bundle/MixedBundleTest.java
```diff
@@ -0,0 +1,282 @@
...
+                    data.extensionWithUsageContext(),
+                    data.extensionWithContactDetail()))
+            .issue(singletonList(data.issue()))
+            .build();
+    assertRoundTrip(
+        MixedBundle.builder().entry(List.of(MixedEntry.builder().resource(r).build())).build());
+  }
+
+  @Test
+  public void patient() {
```
jblefkove-va, line 221: redundant public

us-core-r4/src/test/java/gov/va/api/health/r4/api/bundle/MixedBundleTest.java
```diff
@@ -0,0 +1,282 @@
...
+
+  @Test
+  void immunization() {
+    Immunization r = SampleImmunizations.get().immunization();
+    assertRoundTrip(
+        MixedBundle.builder().entry(List.of(MixedEntry.builder().resource(r).build())).build());
+  }
+
+  @Test
+  public void location() {
```
jblefkove-va, line 169: redundant public

jblefkove-va (dismissed):

us-core-r4/src/test/java/gov/va/api/health/r4/api/bundle/MixedBundleTest.java
```diff
@@ -0,0 +1,242 @@
...
+import gov.va.api.health.r4.api.samples.SamplePractitioners;
+import gov.va.api.health.r4.api.samples.SampleProcedures;
+import gov.va.api.health.r4.api.samples.SampleQuestionnaireResponses;
+import gov.va.api.health.r4.api.samples.SampleQuestionnaires;
+import gov.va.api.health.r4.api.samples.SampleRelatedPersons;
+import gov.va.api.health.r4.api.samples.SampleTerminologyCapabilities;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class MixedBundleTest {
```
jhulbert-va, line 59: Am I missing the test that does a roundTrip on a mixed bundle with different resources?

Also, unimportant nitpick, a ParamaterizedTest here would have been :ok_hand:

us-core-r4/src/test/java/gov/va/api/health/r4/api/bundle/MixedBundleTest.java
```diff
@@ -0,0 +1,316 @@
...
+    ObjectMapper mapper = new JacksonConfig().objectMapper();
+    Patient evilTwin =
+        mapper.readValue(
+            mapper.writerWithDefaultPrettyPrinter().writeValueAsString(pat), Patient.class);
+    assertThat(evilTwin).isEqualTo(SamplePatients.get().patient());
+    MixedBundle bundle =
+        MixedBundle.builder().entry(List.of(MixedEntry.builder().resource(pat).build())).build();
+    String bundleJson = mapper.writerWithDefaultPrettyPrinter().writeValueAsString(bundle);
+    assertThatExceptionOfType(JsonMappingException.class)
+        .isThrownBy(() -> mapper.readValue(bundleJson, MixedBundle.class));
```
jhulbert-va, line 191: Can't this all just become:
```
assertThatExceptionOfType(JsonMappingException.class)
        .isThrownBy(() -> assertRoundTrip(bundleofPatientWithNullResourceType));
```
?

jblefkove-va: I think it's worth breaking out to be more explicit about where the exception is coming from; serializing the bundle is fine, it is deserializing specifically that throws the exception.

(And prior to that, the individual resource itself, even with type missing, can still be round tripped successfully.)

jhulbert-va (approved):
LGTM

us-core-r4/src/test/java/gov/va/api/health/r4/api/bundle/MixedBundleTest.java
jblefkove-va: On second thought I like your version better

jblefkove-va (approved):

us-core-r4/src/test/java/gov/va/api/health/r4/api/bundle/MixedBundleTest.java
```diff
@@ -0,0 +1,136 @@
...
+    assertThatExceptionOfType(JsonMappingException.class).isThrownBy(() -> assertRoundTrip(bundle));
+  }
+
+  @Test
+  @SneakyThrows
+  void exception_unknownResource() {
+    FooResource foo = FooResource.builder().build();
+    // Solo resource can round-trip
+    assertRoundTrip(foo);
+    // Bundled resource requires additional configuration
```
bschofield-va, line 94: We might want to note what the additional configuration is. I do worry about the next resource will get added but not meet the new requirements.

bschofield-va:
We need some sort of document that describes how to add a new resource such that it is compatible with mixed bundles

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Resource.java
```diff
@@ -17,4 +27,58 @@
   String language();
 
   Meta meta();
+
+  public final class ResourceDeserializer extends StdDeserializer<Resource> {
+
+    static final Map<String, Class<? extends Resource>> CLASS_MAP =
+        Map.ofEntries(
```
bschofield-va, line 34: Not a fan of maintaining this map. We will forget to add something to it. Instead of maintaining the map, why not assume any class in this package `gov.va.api.health.r4.api.resources` is a candidate resource.

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Resource.java
```diff
@@ -17,4 +27,58 @@
...
+      super(vc);
+    }
+
+    @Override
+    @SneakyThrows
+    public Resource deserialize(JsonParser jp, DeserializationContext ctxt) {
+      ObjectMapper mapper = (ObjectMapper) jp.getCodec();
+      ObjectNode root = mapper.readTree(jp);
+      String type = root.get("resourceType").asText();
+      Class<? extends Resource> clazz = CLASS_MAP.get(type);
```
bschofield-va, line 77: How about ...
```
clazz = Class.forName(Resource.class.getPackageName() + "." + type);
if (Resource.class.isAssignableFrom(clazz)) { 
  mapper.readValue( ... )
}
throw ...
```

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Resource.java
```diff
@@ -17,4 +27,58 @@
...
+      if (clazz != null) {
+        return mapper.readValue(root.toString(), clazz);
+      }
+      throw new IllegalStateException("Unknown resource type: " + type);
```
bschofield-va, line 81: Is it a bad _state_ to know know the resource type. Perhaps throwing a Jackson deserialization exception is more appropriate.

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Resource.java
jblefkove-va: I like it. This also makes it more explicit that we are not supporting the `gov.va.api.health.r4.api.bundle` package. (Switching from the `@JsonSubTypes` solution to this one allowed us to avoid the problem of `Patient.Bundle` getting turned into `MixedBundle`. That said, nested bundles, e.g. `Patient.Bundle` inside `MixedBundle`, are still not supported. But it would be straightforward to enhance `ResourceDeserializer.deserialize()` later if we needed to add that.)

that-guy-is-it:
With this new approach, it is no longer necessary to document anything, as any class that extends `Resource` in the `resources` package should now work with this. Of note is that since this makes it so every Resource is now serializable without much issue, it might be possible to eventually remove `SimpleResource`. I think, anyway. Outside the scope of the story and this PR, but I figured it was worth bringing it up for the future, @bschofield-va

us-core-r4/src/test/java/gov/va/api/health/r4/api/bundle/MixedBundleTest.java
```diff
@@ -0,0 +1,153 @@
...
```
jblefkove-va, line 111: This comment should be updated

bschofield-va:
> With this new approach, it is no longer necessary to document anything

43 changed files says otherwise. This is clearly a departure from what we used to do, so it's worth a little documentation on how mixed bundles work and what it means to make the next great new resource mixed compatible..

bschofield-va:
I'd like a test case for attempting to deserialize an object in a mixed bundle without `@JsonDeserialize(builder = Whatever.WhateverBuilder.class)`

bschofield-va:
It might be worth updating `RoundTrip.assertRoundTrip(...)` to be `Resource` aware such that when it encounters something that is a resource it perform an additional check of verifying that it is mixed bundled compatible.

I think this is easy to miss.

jblefkove-va:
I'm looking into a solution that does not require adding `@JsonDeserialize` to every class.

@that-guy-is-it To simplify further review, some of the low impact changes in this PR can be split out into their own branch and merged first. Let's make a separate PR just for standardizing the handling of the `resourceType` field. That is:

- Adding `Builder.Default` to any `resourceType` where it is missing
- Adding `defaultString(resourceType, "Bundle")` to every Bundle constructor
- Removing unnecessary overrides of `resourceType` from the unit tests, e.g. `.resourceType("Coverage")` in `SamplePatients`

jblefkove-va:
Also I've added [API-5184](https://vajira.max.gov/browse/API-5184) as a follow-up story to look into replacing `SimpleResource` with `Resource` directly.

us-core-r4/src/test/java/gov/va/api/health/r4/api/bundle/MixedBundleTest.java
```diff
@@ -0,0 +1,147 @@
...
+    var r = SamplePatients.get().patient().resourceType("NotARealResource");
+    var bundle =
+        MixedBundle.builder().entry(List.of(MixedEntry.builder().resource(r).build())).build();
+    assertThatExceptionOfType(JsonMappingException.class).isThrownBy(() -> assertRoundTrip(bundle));
+  }
+
+  @Test
+  void exception_missingClassDeserializer() {
+    MissingClassDeserializerResource r = MissingClassDeserializerResource.builder().build();
+    assertThatExceptionOfType(StackOverflowError.class).isThrownBy(() -> assertRoundTrip(r));
```
bschofield-va, line 89: Expecting a Stack Overflow is not acceptable

bschofield-va (changes requested):
Stack overflow is not ok

jblefkove-va:
@bschofield-va please take another look

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Resource.java
```diff
@@ -17,4 +32,50 @@
...
+      super(vc);
+    }
+
+    @Override
+    @SneakyThrows
+    public Resource deserialize(JsonParser jp, DeserializationContext ctxt) {
+      ObjectMapper mapper = (ObjectMapper) jp.getCodec();
+      ObjectNode root = mapper.readTree(jp);
+      String type = root.get("resourceType").asText();
+      if (type.equals("SimpleResource")) {
```
bschofield-va, line 51: Why is this special? That is, why do we specifically look for simple resource? What makes a simple resource so different from every other resource?

jblefkove-va: it's in a different package

bschofield-va: Perhaps a comment to explain why so future us still understands?

bschofield-va (dismissed):

jblefkove-va (approved):

PR #118, opened 2021-02-18
Added missing types to Extension

that-guy-is-it:
https://vajira.max.gov/browse/API-3303


jblefkove-va:
while you are in here, `url` is a required field

jblefkove-va:
also fix the ordering of fields: should be id, then extension, then url

jblefkove-va:
`valueSampledData` is missing from the `zeroOrOneOf` annotation

us-core-r4/src/main/java/gov/va/api/health/r4/api/datatypes/Count.java
```diff
@@ -0,0 +1,43 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@Schema(description = "http://hl7.org/fhir/R4/datatypes.html#Count")
+public class Count implements Element {
```
jblefkove-va, line 24: The FHIR spec defines `Count` and `Distance` datatypes, but indicates that they are not used anywhere (yet). I'm inclined to do the same thing in `Extension` that we did with `Parameters`: model count and distance as `Quantity` rather than defining dedicated classes for them.

jblefkove-va (approved):

PR #126, opened 2021-06-07
update links to documentation

ilaflamme-va:
https://vajira.max.gov/browse/API-7607

After reviewing our model, and the data-mappings spreadsheet . it looks like we can build a compliant Organization from Vista.

Changes include fixing a broken link to the specification.

### DIFF
--- a/us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
+++ b/us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
@@ -46,7 +46,8 @@ import org.apache.commons.lang3.StringUtils;
     isGetterVisibility = JsonAutoDetect.Visibility.NONE)
 @Schema(
     description =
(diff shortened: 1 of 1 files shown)


us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
```diff
@@ -46,7 +46,8 @@
     isGetterVisibility = JsonAutoDetect.Visibility.NONE)
 @Schema(
     description =
-        "https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-organization.html",
+        "The Organization resource is compliant with both https://www.hl7.org/fhir/organization.html" +
```
jhulbert-va, line 49: Meh. If it's compliant with the second, it's compliant with the first. Not sure we need to state that explicitly though.

eclendenning-va (approved):

PR #127, opened 2021-06-07
API-7729 appointment swagger

jblefkove-va:
https://vajira.max.gov/browse/API-7729

### DIFF
--- a/us-core-r4/src/main/java/gov/va/api/health/r4/api/AppointmentApi.java
+++ b/us-core-r4/src/main/java/gov/va/api/health/r4/api/AppointmentApi.java
@@ -124,6 +124,6 @@ public interface AppointmentApi {
       @Parameter(
               in = ParameterIn.QUERY,
               name = "_sort",
-              description = "Order in which to return results. Date is supported.")
+              description = "Parameters by which to sort results.")
           String sort);
 }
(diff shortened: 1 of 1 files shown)


ilaflamme-va (dismissed):

us-core-r4/src/main/java/gov/va/api/health/r4/api/AppointmentApi.java
```diff
@@ -120,5 +120,10 @@ Appointment appointmentRead(
                   "The number of resources that should be returned in a single page."
                       + " The maximum count size is 100.")
           @DefaultValue("30")
-          int count);
+          int count,
+      @Parameter(
+              in = ParameterIn.QUERY,
+              name = "_sort",
+              description = "Order in which to return results. Date is supported.")
```
bschofield-va, line 127: Does this mean only Date is supported?

jblefkove-va: Yes, I didn't implement any other searches, date was the only one requested right now.

jhulbert-va:
```
Try to sort the results using the given parameters.
```
What about something like that? Kinda implies that it can't be sorted by all of them, but in a way that we don't have to keep a running list.

jblefkove-va: how about `comma-separated parameters by which to sort results`?

jhulbert-va (approved):

PR #128, opened 2021-06-07
Add identifier search param to practitioner swagger

apeirats-va:
Practitioner identifier search allows for NPI searches.

### DIFF
--- a/us-core-r4/src/main/java/gov/va/api/health/r4/api/PractitionerApi.java
+++ b/us-core-r4/src/main/java/gov/va/api/health/r4/api/PractitionerApi.java
@@ -88,7 +88,7 @@ public interface PractitionerApi {
       @Parameter(
               in = ParameterIn.PATH,
               name = "identifier",
-              description = "An identifier uniquely identifies one object/resource.")
+              description = "A unique identifier for a practitioner within a given system.")
           String identifier,
       @Parameter(
               in = ParameterIn.QUERY,
--- a/us-core-r4/src/main/java/gov/va/api/health/r4/api/PractitionerApi.java
+++ b/us-core-r4/src/main/java/gov/va/api/health/r4/api/PractitionerApi.java
@@ -88,11 +88,7 @@ public interface PractitionerApi {
       @Parameter(
               in = ParameterIn.PATH,
               name = "identifier",
-              description =
-                  "The logical identifier of the resource."
-                      + " Once assigned, this value never changes."
-                      + " For Practitioners this identifier is the"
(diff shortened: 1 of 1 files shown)


us-core-r4/src/main/java/gov/va/api/health/r4/api/PractitionerApi.java
```diff
@@ -85,6 +85,15 @@ Practitioner practitionerRead(
...
                   "The logical id of the resource. Once assigned, this value never changes.")
           String id,
+      @Parameter(
+              in = ParameterIn.PATH,
+              name = "identifier",
+              description =
+                  "The logical identifier of the resource."
+                      + " Once assigned, this value never changes."
+                      + " For Practitioners this identifier is the"
+                      + " National Provider Identifier (NPI).")
```
jhulbert-va, line 95: This isn't technically true. If given an I2 id, it's the logical identifier, but an npi isn't the logical identifier. In addition, the identifier parameter is a token and therefore can be used to search for a practitioner by any system of identifiers we allow. 
In contrast, `_id` only supports the logical identifier (I2 id).

jhulbert-va:
```
An id that can be used to identify a practitioner (e.g. the resource's logical id, the practitioner's NPI).
```
Something like that maybe?

jblefkove-va: `A unique identifier for a practitioner within a given system`?

jhulbert-va: Identifier's don't necessarily need to be unique do they?

jhulbert-va: That looks fine to me then.

apeirats-va: I should note that the rest of the `identifier` descriptions use the same "The logical id of the resource..." line.

jhulbert-va (dismissed):

us-core-r4/src/main/java/gov/va/api/health/r4/api/PractitionerApi.java
jhulbert-va: We may want to double check those, but the majority of the other resources only allow identifier searching using the logical ID, so that wouldn't _technically_ be incorrect.

PR #130, opened 2021-06-09
API-7694: remove paging documentation for R4 Observation search

jabando-va:


ilaflamme-va (approved):

eclendenning-va (changes requested):
DQ uses paging

eclendenning-va:
Removing paging from VFQ Observation openapi needs to be done via the conformance unifier.

PR #133, opened 2021-06-14
API-7988: STU3 model discrepancies

apeirats-va:
Some STU3 integration tests are failing on DQ due to a schema validation error. Need to update this schema to match the base STU3 spec before IT failures can be addressed.

DQ PR: https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/755

### DIFF
--- a/argonaut-stu3/src/main/java/gov/va/api/health/stu3/api/resources/Location.java
+++ b/argonaut-stu3/src/main/java/gov/va/api/health/stu3/api/resources/Location.java
@@ -196,9 +196,9 @@ public class Location implements DomainResource {
 
     @Valid List<Extension> modifierExtension;
 
-    @NotNull double longitude;
+    @NotNull Double longitude;
 
-    @NotNull double latitude;
+    @NotNull Double latitude;
 
     double altitude;
   }
--- a/argonaut-stu3/src/main/java/gov/va/api/health/stu3/api/resources/Location.java
+++ b/argonaut-stu3/src/main/java/gov/va/api/health/stu3/api/resources/Location.java
@@ -196,10 +196,10 @@ public class Location implements DomainResource {
 
     @Valid List<Extension> modifierExtension;
 
-    @NotNull Double longitude;
+    @NotNull double longitude;
 
-    @NotNull Double latitude;
+    @NotNull double latitude;
 
-    double altitude;
+    Double altitude;
   }
 }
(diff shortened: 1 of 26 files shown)


jblefkove-va:
@bschofield-va does this need to be a major version change? (Normally I would say so, but DQ is the only consumer here and this is effectively a bugfix against the spec)

jblefkove-va:
Since we're updating stu3 models, some other discrepancies:
1. practitioner.active should be Boolean
2. practitioner.name should be list
3. practitioner.practitionerRole class appears to be unused copy+paste from dstu2
4. location.position.altitude should be Double
5. endpoint.payloadMimeType should be a list

argonaut-stu3/src/main/java/gov/va/api/health/stu3/api/resources/Practitioner.java
```diff
@@ -222,7 +222,7 @@ public Bundle(
   @NoArgsConstructor(access = AccessLevel.PRIVATE)
   @AllArgsConstructor
   @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
-  public static class PractitionerRole implements BackboneElement {
+  public static class PractitionerRoleDeleteMe implements BackboneElement {
```
jblefkove-va, line 225: If we're confident this isn't being used anywhere (DQ would be the only place), let's just delete it.

jblefkove-va:
Let's do a major version rev

argonaut-stu3/src/main/java/gov/va/api/health/stu3/api/resources/Location.java
```diff
@@ -196,9 +196,9 @@ public Entry(
 
     @Valid List<Extension> modifierExtension;
 
-    @NotNull double longitude;
+    @NotNull Double longitude;
 
-    @NotNull double latitude;
+    @NotNull Double latitude;
```
jblefkove-va, line 201: This is reversed. `longitude` and `latitude` should be primitive double since they are required fields, while `altitude` is optional.

argonaut-stu3/src/main/java/gov/va/api/health/stu3/api/resources/Practitioner.java
apeirats-va: Goof on my part. Meant to delete after renaming.

jblefkove-va (approved):

PR #142, opened 2021-08-13
swagger practitioner exaample

ilaflamme-va:


us-core-r4-examples/src/main/java/gov/va/api/health/r4/api/swaggerexamples/SwaggerPractitoner.java
```diff
@@ -0,0 +1,88 @@
+package gov.va.api.health.r4.api.swaggerexamples;
```
bschofield-va, line 1: This one looks good... the DSTU2 sample is no bueno, tho.

ilaflamme-va: I've updated the DSTU2 sample as well, now.

bschofield-va (approved):

PR #145, opened 2021-09-10
insurance plan

eclendenning-va:
swagger in follow on PR


us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/InsurancePlan.java
```diff
@@ -0,0 +1,366 @@
...
+
+  List<String> alias;
+
+  @Valid Period period;
+
+  @Valid Reference ownedBy;
+
+  @Valid Reference administeredBy;
+
+  @Valid Reference coverageArea;
```
bschofield-va, line 98: is list

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/InsurancePlan.java
```diff
@@ -0,0 +1,366 @@
...
+      @Pattern(regexp = Fhir.ID)
+      String id;
+
+      @Valid List<Extension> extension;
+
+      @Valid List<Extension> modifierExtension;
+
+      @Valid CodeableConcept type;
+
+      @PositiveOrZero Integer groupSize;
```
bschofield-va, line 295: This should be a positive int:
> Any positive integer in the range 1..2,147,483,647

bschofield-va (approved):

PR #164, opened 2022-01-25
Api 12334 update immunization with manufacturer notes

taylorrassmannlibertyits:


us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Immunization.java
```diff
@@ -330,4 +330,4 @@ public Entry(
 
     Boolean reported;
   }
-}
+}
```
jblefkove-va, line 333: sus

jblefkove-va (approved):

PR #165, opened 2022-01-28
API-000-domain-resources

jhulbert-va:
# [API-000](https://vajira.max.gov/browse/API-000)

I upgraded the major version because I had to fix the practitioner extension/modifierExtension fields which could affect any applicatino using the current version of that resource.


eclendenning-va (approved):

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Practitioner.java
```diff
@@ -71,9 +71,9 @@
 
   @Valid List<Resource> contained;
 
-  @Valid List<Extension> extensions;
+  @Valid List<Extension> extension;
```
bschofield-va, line 74: good catch

PR #166, opened 2022-02-01
allow contained resources to be deserialized without module

eclendenning-va:


jhulbert-va (dismissed):

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Organization.java
```diff
@@ -154,7 +156,6 @@ private boolean isValidIdentifierNpiSlice() {
   @NoArgsConstructor
   @EqualsAndHashCode(callSuper = true)
   @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
```
bschofield-va, line 158: ^ is `@JsonAutoDetect` still needed?

bschofield-va: This annotation appears in a few othre classes, too

bschofield-va:
I guess I was expecting to see some sort of unit test that proved this worked.

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Resource.java
```diff
@@ -19,9 +20,9 @@
 import javax.validation.constraints.Pattern;
 import lombok.SneakyThrows;
 
+@JsonDeserialize(using = Resource.ResourceDeserializer.class)
```
jblefkove-va, line 23: Since we configure this behavior directly on the class now, do we still need `ResourceModule`?

jblefkove-va:
I see a bunch of unit test failures with `StackOverflowException` when I run them out of my IDE... local Maven build works, but still sus

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/AllergyIntolerance.java
```diff
@@ -194,9 +196,9 @@ public Bundle(
   @EqualsAndHashCode(callSuper = true)
   @Schema(name = "AllergyIntoleranceEntry")
   @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
-  @JsonDeserialize(builder = AllergyIntolerance.Entry.EntryBuilder.class)
   public static final class Entry extends AbstractEntry<AllergyIntolerance>
       implements AsList<Entry> {
+    @Jacksonized
```
jblefkove-va, line 201: Should we be applying `@Jacksonized` to the builders as well?

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Resource.java
eclendenning-va: If we remove the module we don't need a customized mapper. But your stack overflows are worrisome, I am not getting them in my IDE or through local maven.

jhulbert-va (dismissed):

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Resource.java
```diff
@@ -28,48 +29,56 @@
...
-  public final class ResourceModule extends SimpleModule {
-    public ResourceModule() {
-      addDeserializer(Resource.class, new ResourceDeserializer());
+    @Override
+    @SneakyThrows
+    public JavaType typeFromId(DatabindContext context, String id) {
+      try {
+        return context.constructType(Class.forName(id));
+      } catch (ClassNotFoundException e) {
+        if (id.endsWith("$Bundle")) {
```
jblefkove-va, line 77: This is protection for classes that do not define an internal bundle type, like `OperationOutcome` or `Parameters`

bschofield-va: ^ perhaps this is worth a comment in the code for future us

us-core-r4/src/test/java/gov/va/api/health/r4/api/RoundTrip.java
```diff
@@ -11,20 +12,26 @@
 import lombok.SneakyThrows;
 
 public class RoundTrip {
+  private static final ObjectMapper MAPPER = JacksonConfig.createMapper();
```
bschofield-va, line 15: what are the odds this will work with `new ObjectMapper()` vs. JacksonConfig?

jblefkove-va: Checking...

jblefkove-va: updated

jblefkove-va:
I think this is good-to-go now

jblefkove-va:
After integration testing with DQ, there is a deficiency with this approach. Will explain in more detail tomorrow.

jblefkove-va:
I've moved the missing annotations into their own PR: https://github.com/department-of-veterans-affairs/health-apis-fhir-resources/pull/168

jblefkove-va:
While testing these changes against DQ, I observed basic deserialization of resources not working. I forgot to add `include = JsonTypeInfo.As.EXISTING_PROPERTY` on the `JsonTypeInfo` annotation. All of the `roundTrip` tests were coincidentally working because serialization runs first; a second `resourceType` property was being created with the output from the `idFromValue(Object value)` method, which is a fully-qualified class name, allowing deserialization to work.

However, when just deserializing, only the `typeFromId(DatabindContext context, String id)` method is run... unless I'm missing something, that method only has access to the value of `resourceType`, but cannot inspect the full object. So that rather limits the options for deserializing `MixedBundle` vs inner bundle types.

I've added unit tests in https://github.com/department-of-veterans-affairs/health-apis-fhir-resources/pull/168 to catch the deserialization problem, but the other issue requires further thought.

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Resource.java
```diff
@@ -28,48 +29,33 @@
 
   Meta meta();
 
-  public final class ResourceDeserializer extends StdDeserializer<Resource> {
-    public ResourceDeserializer() {
-      this(null);
-    }
+  String resourceType();
```
jhulbert-va, line 32: Should `resourceType` be defined in `IsResource`?

jblefkove-va: Yeah

jblefkove-va:
DQ upgrade: https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/870

jblefkove-va:
Ok, we have a working, but slightly gross solution. All typed bundle classes are annotated with `@JsonTypeInfo`. If you explicitly specify a typed Bundle class during deserialization, the `@JsonTypeInfo` annotation on that class will override the behavior in `ResourceTypeIdResolver`. In any other case, a bundle will be deserialized as `MixedBundle`. Not ideal, but DQ looks good with this solution, I think we can proceed.

bschofield-va (dismissed):

bschofield-va (approved):

PR #168, opened 2022-02-04
API-13103 add missing annotations for vanilla object mapper

jblefkove-va:
- pull out changes from https://github.com/department-of-veterans-affairs/health-apis-fhir-resources/pull/166 to add missing annotations to certain files
- additional tests for `MixedBundle`


us-core-r4/src/test/java/gov/va/api/health/r4/api/RoundTrip.java
```diff
@@ -11,20 +11,22 @@
 import lombok.SneakyThrows;
 
 public class RoundTrip {
+  public static final ObjectMapper MAPPER =
+      new ObjectMapper().registerModule(new Resource.ResourceModule());
```
jhulbert-va, line 15: Is there a future PR coming that does not register the module?

jhulbert-va (approved):

us-core-r4/src/test/java/gov/va/api/health/r4/api/RoundTrip.java
jblefkove-va: Right

PR #170, opened 2022-03-02
API-13843 add R4 DeviceRequest

jblefkove-va:


us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/DeviceRequest.java
```diff
@@ -0,0 +1,272 @@
...
+
+  Status status;
+
+  @NotNull Intent intent;
+
+  Priority priority;
+
+  @Valid Reference codeReference;
+
+  @Valid CodeableConcept codeCodeableConcept;
```
agarcia-va, line 104: Shouldn't this and `codeReference` both be `@NotNull`? Both are `1..1` in the FHIR docs.

jblefkove-va: As a group, exactly one of `codeReference` and `codeCodeableConcept` must be non-null, and that is validated at the class-level. Each individual field could still be null, though.

taylorrassmannlibertyits (approved):

PR #171, opened 2022-03-14
New HasTargetPatient interface for FHIR Common and implementations for R4 resources

bschofield-va:


fhir-common/src/main/java/gov/va/api/health/fhir/api/HasTargetPatient.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.regex.Pattern;
+
+/** Indicates that an item that targets a single specific patient, like an Observation record. */
+public interface HasTargetPatient {
+
+  /** Regex to match whatever/Patient/123 type reference values. */
+  Pattern PATIENT_REFERENCE = Pattern.compile("(^|.*/)Patient/[^/]+");
+
+  /** Return true if the given reference is referring to a patient resource. */
+  static boolean isPatientReference(IsReference reference) {
```
jhulbert-va, line 13: I wonder if `IsReference` should also know about `.type` so it can check that. This currently won't work for contained references.

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Appointment.java
```diff
@@ -112,6 +115,15 @@
 
   @Valid List<Period> requestedPeriod;
 
+  @Override
+  public IsReference targetPatientId() {
+    return Safe.stream(participant())
+        .map(Participant::actor)
+        .filter(HasTargetPatient::isPatientReference)
+        .findFirst()
```
jhulbert-va, line 123: Couldn't this lead to us just ignoring an patient reference that should result in a patient matching failure?

fhir-common/src/main/java/gov/va/api/health/fhir/api/HasTargetPatient.java
bschofield-va: Maybe, but we don't populate type most of the time

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Appointment.java
bschofield-va: if, for example an appointment was for _two_ patients?

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/QuestionnaireResponse.java
```diff
@@ -97,6 +100,11 @@
 
   @Valid List<Item> item;
 
+  @Override
+  public IsReference targetPatientId() {
+    return subject();
```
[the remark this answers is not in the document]
bschofield-va: Let me do some thinking and what I might be able to do here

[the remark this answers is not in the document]
bschofield-va: I made check a few different possible references and pick the first patient it finds

jblefkove-va (approved):

PR #172, opened 2022-03-16
Update medication dispense and request swagger examples

msilverman-va:
Co-authored-by: Joshua Miller <jmiller@rise8.us>
Co-authored-by: Thomas Reynolds <treynolds@rise8.us>


jnmiller-va:
@kivarsson-va did we want to update to ready for review?

kivarsson-va:
@jnmiller-va whoops yeah meant to come back when checks were done

us-core-r4-examples/src/main/java/gov/va/api/health/r4/api/swaggerexamples/SwaggerMedicationDispense.java
```diff
@@ -25,42 +27,65 @@ public class SwaggerMedicationDispense {
...
+                        .build()
+                        .asList())
+                .build()
+                .asList())
         .authorizingPrescription(
             Reference.builder()
                 .reference(
                     "https://sandbox-api.va.gov/services/fhir/v0/r4/MedicationRequest/I2-AOV4FXGQLPIXGZPTMTWY7Y7KJ4000000")
-                .display("RX")
+                .id("I2-AOV4FXGQLPIXGZPTMTWY7Y7KJ4000000")
```
bschofield-va, line 78: We don't populate `id` of `Reference`. Please remove.

kivarsson-va:
@jnmiller-va or @msilverman-va can correct me if wrong but i think we were asked to update the swagger to better match what the API was serving up when we were going to prod? if y'all don't want this feel free to chuck it. i don't think we're extremely attached to it @bschofield-va

bschofield-va:
I'm afraid that sample falls short of the goal. It doesn't actually look like what will be produced, e.g.
```
{
  "resourceType": "MedicationDispense",
  "id": "I3-Gb8CL0jU5H28XaefbHqObnqlEf4hfOZhOcf8m4eMI93bKt0quIh75z",
  "meta": {
    "source": "673"
  },
  "extension": [
    {
      "url": "http://hl7.org/fhir/StructureDefinition/medicationdispense-refillsRemaining",
      "valueInteger": 2
    }
  ],
  "status": "completed",
  "medicationCodeableConcept": {
    "coding": [
      {
        "system": "https://www.pbm.va.gov/nationalformulary.asp",
        "code": "CN103",
        "display": "NON-OPIOID ANALGESICS"
      }
    ],
    "text": "ACETAMINOPHEN 500MG TAB"
  },
  "subject": {
    "reference": "https://sandbox-api.va.gov/services/clinical-fhir/v0/r4/Patient/1011537977V693883"
  },
  "location": {
    "display": "TAMPA (JAH VAH)"
  },
  "authorizingPrescription": [
    {
      "reference": "https://sandbox-api.va.gov/services/clinical-fhir/v0/hcs/673/r4/MedicationRequest/I3-GTBBRHbarRNLFP8vUaCBY64wGpN0TxB7HWsJxJNbhuu"
    }
  ],
  "quantity": {
    "value": 90
  },
  "daysSupply": {
    "value": 90,
    "unit": "day",
    "system": "http://unitsofmeasure.org",
    "code": "d"
  },
  "whenPrepared": "2008-07-28T00:00:00Z",
  "whenHandedOver": "2008-07-28T00:00:00Z",
  "destination": {
    "display": "WINDOW"
  },
  "dosageInstruction": [
    {
      "text": "TAKE ONE TABLET BY MOUTH EVERY DAY",
      "doseAndRate": [
        {
          "doseQuantity": {
            "value": 500,
            "unit": "MG"
          }
        }
      ]
    }
  ]
}
```

kivarsson-va:
@bschofield-va perhaps we misunderstood the goal? this looks pretty much the same, save for the exact IDs that are used (and the `id` field which i admittedly overlooked when making my edits). what exactly are you looking for this to be?

bschofield-va:
Some URLs reference `services/fhir/v0` which is the Veteran Health API.
We've had defects written against our APIs before because the sample IDs are _unusable_ in the Sandbox and we've had to make them real sample IDs.

kivarsson-va:
okay, we can update it then with that context. when we had originally made these changes we were still waiting on access to everything in sandbox and were just approximating values. no problem.

kivarsson-va:
@bschofield-va now that you mention the vet health API -- [this same issue](https://lighthouseva.slack.com/archives/CTTEGCQ01/p1649175365214549?thread_ts=1648142423.451679&cid=CTTEGCQ01) exists for dispense, should we just forget about this for now?

bschofield-va:
You definitely can't change MedicateRequest as it would break VHAPI.
I think I mentioned we had this this swagger example problem (1 example, n APIs) once before. But if not, we have this problem 😭. We're working on a new solution to generate examples based on the live API ... I think @jblefkove-va is running a spike on that this sprint. 

I don't have strong feelings on whether this change goes in or not, but if it's going in MedicationDispense should be correct for CHAPI.

kivarsson-va:
ah right but VHAPI doesn't have MedicationDispense (yet), i was forgetting. we might as well make dispense up to date since it's only relevant for CHAPI anyway -- we'll get the issues with this PR fixed 👍

bschofield-va (approved):

jblefkove-va (approved):

PR #175, opened 2022-03-23
API 13930 patient centric resources part 9

bschofield-va:
- New @HasTargetPatientQueryParameter and @DoesNotHaveTargetPatientQueryParameter annotations for Resource classes
- Round trip tests now verify the R4 resource has declared Has/DoesNotHaveTargetPatient and Has/DoesNotHaveTargetPatientQueryParameter


eclendenning-va (approved):

us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Endpoint.java
```diff
@@ -47,6 +48,7 @@
     description = "http://www.fhir.org/guides/argonaut/pd/StructureDefinition-argo-endpoint.html",
     example =
         "${r4.endpoint:gov.va.api.health.r4.api.swaggerexamples." + "SwaggerEndpoint#endpoint}")
+@DoesNotHaveTargetPatientQueryParameter
```
jhulbert-va, line 51: We do define a patient query parameter for Endpoint.

bschofield-va: yes, but not standard

jhulbert-va (approved):

jblefkove-va (approved):

PR #194, opened 2022-04-21
API-000-optional-paging

jhulbert-va:
# [API-000](https://vajira.max.gov/browse/API-000)


eclendenning-va (approved):

test-support/src/main/java/gov/va/api/health/fhir/testsupport/ResourceVerifier.java
```diff
@@ -132,7 +134,7 @@ public final <T> TestCase<T> test(
 
   /** Verify a single test case. */
   public final <T> T verify(TestCase<T> tc) {
-    if (!tc.isPost()) {
+    if (!tc.isPost() && assertPagingParameters()) {
```
jblefkove-va, line 137: Would this check be more appropriate at the beginning of `assertPagingParameterBounds()`? (And would `assertPagingParameterBounds` also then make more sense for the field name?)

PR #197, opened 2022-04-28
API-15528 Added reason code example to swagger

taylorrassmannlibertyits:


us-core-r4-examples/src/main/java/gov/va/api/health/r4/api/swaggerexamples/SwaggerDeviceRequest.java
```diff
@@ -31,6 +32,17 @@ public static DeviceRequest deviceRequest() {
                     "https://sandbox-api.va.gov/services/fhir/v0/r4/Practitioner/I2-HRJI2MVST2IQSPR7U5SACWIWZA000000")
                 .display("DR. JANE460 DOE922 MD")
                 .build())
+        .reasonCode(
+            CodeableConcept.builder()
+                .coding(
+                    Coding.builder()
+                        .version("10")
```
jblefkove-va, line 39: `system` should also be specified

jblefkove-va (approved):

PR #203, opened 2022-06-17
print full URL in test case label

jblefkove-va:


test-support/src/main/java/gov/va/api/health/fhir/testsupport/ResourceVerifier.java
```diff
@@ -166,6 +168,8 @@ public final void verifyAll(TestCase<?>... testCases) {
 
     String path;
```
taylorrassmannlibertyits, line 169: Can path be removed?

jblefkove-va: No, the short path is still used in other places

taylorrassmannlibertyits (approved):

PR #211, opened 2022-08-12
API 18190 better deserialization support

bschofield-va:
- Remove SimpleResource and replace with plain Resource
- Major version 12
- R4-DSTU2 Condition now ignored contained resources


jblefkove-va (dismissed):

jblefkove-va (approved):

deidentification-core/pom.xml
```diff
@@ -4,10 +4,10 @@
   <parent>
     <artifactId>fhir-resources-parent</artifactId>
     <groupId>gov.va.api.health</groupId>
-    <version>11.0.21-SNAPSHOT</version>
+    <version>12.0.0-SNAPSHOT</version>
   </parent>
   <artifactId>deidentification-core</artifactId>
-  <version>11.0.21-SNAPSHOT</version>
+  <version>12.0.0-SNAPSHOT</version>
```
jhulbert-va, line 10: We shouldn't need this I don't think.

PR #221, opened 2022-08-25
API-18967-find-matching-changeable-params

jhulbert-va:
# [API-18967](https://vajira.max.gov/browse/API-18967)

The new method will be used in yanartas to determine whether or not to use the target icn for doppelgagners.


eclendenning-va (approved):

jblefkove-va (approved):
:shipit:

bcastillomartel-va (approved):

PR #222, opened 2022-08-26
Add Binary Model

KevinCranmer:
https://vajira.max.gov/browse/API-18875


us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Binary.java
```diff
@@ -0,0 +1,143 @@
...
+  @Valid Meta meta;
+
+  @Pattern(regexp = Fhir.URI)
+  String implicitRules;
+
+  @Pattern(regexp = Fhir.CODE)
+  String language;
+
+  // Ancestor -- DomainResource
+  @Valid Narrative text;
```
jblefkove-va, line 62: `Binary` is not `DomainResource`

jblefkove-va (approved):

PR #231, opened 2022-09-06
API-19466-datetime-converter

jhulbert-va:
# [API-19466](https://vajira.max.gov/browse/API-19466)


fhir-resource-conversion/src/test/java/gov/va/api/health/fhir/conversion/ConvertersTest.java
```diff
@@ -25,6 +27,21 @@ void asList() {
...
   }
 
+  @ParameterizedTest
+  @CsvSource(
+      nullValues = "null",
+      textBlock =
+          """
+    null,null
+    ,null
+    2020-01-20T03:00:00Z,2020-01-20
```
jblefkove-va, line 37: do you also want to test another time-zone?

jblefkove-va (approved):

PR #249, opened 2022-10-25
Api 18514 add medication converter

KevinCranmer:
https://vajira.max.gov/browse/API-18514


fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/datatypes/RatioConverter.java
```diff
@@ -0,0 +1,13 @@
+package gov.va.api.health.fhir.conversion.r4todstu2.datatypes;
+
+import gov.va.api.health.fhir.conversion.Converter;
+import gov.va.api.health.fhir.conversion.r4todstu2.config.ElementConverterConfig;
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+@Mapper(config = ElementConverterConfig.class)
+public abstract class RatioConverter
```
jhulbert-va, line 9: Was this needed (i.e. did it cause breakages when it didn't exist)? In general mapstruct is pretty good about generating sub-mappings for you. If it was needed, make sure you add it to the ResourceConverterConfig.

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationConverter.java
```diff
@@ -0,0 +1,76 @@
...
+    implements ResourceConverter<
+        gov.va.api.health.r4.api.resources.Medication,
+        gov.va.api.health.dstu2.api.resources.Medication> {
+  public static final MedicationConverter INSTANCE = Mappers.getMapper(MedicationConverter.class);
+
+  @Mapping(
+      target = "product",
+      expression =
+          "java(convertProduct(source.form(), source.ingredient(),"
+              + "source.amount(), source.batch(), source.identifier()))")
```
jhulbert-va, line 29: In general, I like to format these in a way that reads like how jarranger would format. That is to say I think that parameter from the method should be on it's own line.

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationConverter.java
```diff
@@ -0,0 +1,76 @@
...
+  @Mapping(target = "contained", ignore = true)
+  @Mapping(target = "brand", ignore = true)
+  @Mapping(target = "medicationPackage", ignore = true)
```
jhulbert-va, line 32: These don't map to DSTU2?

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationConverter.java
```diff
@@ -0,0 +1,76 @@
...
+      gov.va.api.health.r4.api.datatypes.Ratio amount,
+      gov.va.api.health.r4.api.resources.Medication.Batch batch,
+      List<gov.va.api.health.r4.api.datatypes.Identifier> identifiers) {
+    if (form == null
+        && ingredient == null
+        && amount == null
+        && batch == null
+        && identifiers == null) {
+      return null;
+    }
```
jhulbert-va, line 49: You'll want to perform checks for empty on the lists.

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationConverter.java
KevinCranmer: Looks like it can map from r4 extension to dstu2 medicationPackage:
http://hl7.org/fhir/medication-version-maps.html
http://hl7.org/fhir/STU3/medication-version-maps.html

I'll add that functionality

KevinCranmer: I added conversion for the medicationPackage but seems sus. Let me know what you think about it/if it will ever be used.

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/datatypes/RatioConverter.java
KevinCranmer: I want to use it inside `convertProduct()` is there a way I can leverage the generated sub-mappings similar to:
`RatioConverter.INSTANCE.convert(amount)`
?

KevinCranmer:
Struggling to understand the null pointer exception in the build. I don't get the error locally doing an `mci`. But I get a different null pointer exception when running the test individually through IntelliJ. It's like the `MedicationConverter` is not hooked up to the `ResourceConverterConfig` but I have the annotation for it...

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationConverter.java
```diff
@@ -0,0 +1,146 @@
...
+import org.mapstruct.Mapping;
+import org.mapstruct.Named;
+import org.mapstruct.factory.Mappers;
+
+@Mapper(config = ResourceConverterConfig.class)
+public abstract class MedicationConverter
+    implements ResourceConverter<
+        gov.va.api.health.r4.api.resources.Medication,
+        gov.va.api.health.dstu2.api.resources.Medication> {
+  public static final MedicationConverter INSTANCE = Mappers.getMapper(MedicationConverter.class);
```
jblefkove-va, line 23: Do we still need `INSTANCE` on the resource converters?

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationConverter.java
```diff
@@ -0,0 +1,146 @@
...
+  @Named("convertExtension")
+  List<gov.va.api.health.dstu2.api.elements.Extension> convertExtension(
+      List<gov.va.api.health.r4.api.elements.Extension> extensions) {
+    if (extensions == null) {
+      return null;
+    }
+    List<gov.va.api.health.dstu2.api.elements.Extension> dstu2Extensions = new ArrayList<>();
+    for (gov.va.api.health.r4.api.elements.Extension e : extensions) {
+      if (!e.url().contains("StructureDefinition/extension-Medication.package.container")
+          && !e.url().contains("StructureDefinition/extension-Medication.package.content")) {
```
jblefkove-va, line 52: Why are these extensions filtered?

KevinCranmer: My thought was that a dstu2 Medication does not have these extensions. Rather, the dstu2Medication medicationPackage is converted into extensions going to r4. So when converting back to dstu2, instead of duplicating data in the extensions and medicationPackage, I removed the extensions.

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationConverter.java
KevinCranmer: I believe so. Other Converters don't use them but maybe cerner does somehow? @jhulbert-va

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationConverter.java
```diff
@@ -0,0 +1,166 @@
...
+  gov.va.api.health.dstu2.api.resources.Medication.MedicationPackage convertMedicationPackage(
+      List<gov.va.api.health.r4.api.elements.Extension> extensions) {
+    if (extensions == null
+        || extensions.stream()
+            .noneMatch(
+                e ->
+                    e.url().contains("StructureDefinition/extension-Medication.package.container")
+                        || e.url()
+                            .contains(
+                                "StructureDefinition/extension-Medication.package.content"))) {
```
jhulbert-va, line 106: This can be used in the above method as well (Streams are fun!) why not make a method - isMedicationPackageExtension - that takes an extension and returns a boolean. Then you could use streams to filter/map above and make this if statement way more readable.

jhulbert-va: Ooo! With that you could even:
```
var medPackageExtensions = Safe.stream(extensions).filter(this::isMedicationPackageExtension).toList();
if (medPackageExtensions.isEmpty()) {
  return null;
}
```

^ The you would only iterate the _full_ list once.

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationConverter.java
```diff
@@ -0,0 +1,166 @@
...
+      List<gov.va.api.health.r4.api.datatypes.Identifier> identifiers) {
+    if (form == null
+        && (ingredient == null || ingredient.isEmpty())
+        && amount == null
+        && batch == null
+        && (identifiers == null || identifiers.isEmpty())) {
+      return null;
+    }
+    return gov.va.api.health.dstu2.api.resources.Medication.Product.builder()
+        .id(Converters.INSTANCE.firstElementAndConvertOrDie(identifiers, Identifier::id))
```
jhulbert-va, line 141: Identifier id? Or should this be using Identifier.value?

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationConverter.java
```diff
@@ -0,0 +1,166 @@
...
+                        .amount(FieldConverter.INSTANCE.convert(amount))
+                        .build()))
+        .batch(
+            batch == null
+                ? null
+                : List.of(
+                    gov.va.api.health.dstu2.api.resources.Medication.Batch.builder()
+                        .lotNumber(batch.lotNumber())
+                        .expirationDate(batch.expirationDate())
+                        .build()))
```
jhulbert-va, line 158: Why not create an abstract method to convert batch and then just call that here?

fhir-resource-conversion/src/test/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationConverterTest.java
```diff
@@ -0,0 +1,352 @@
...
+                null,
+                List.of(
+                    MedicationSamples.R4.extension("Extension/e1", e -> e.valueString("test"))))));
+  }
+
+  static Stream<Arguments> convertProduct() {
+    return Stream.of(
+        Arguments.of(null, null, null, null, null, null),
+        Arguments.of(
+            product(dstu2Form(), dstu2Ingredients(dstu2Amount()), dstu2Batch(), dstu2Ids()),
```
jhulbert-va, line 199: Instead of creating methods whose singular function is to call methods you've defined in samples, why not create a variable at the top `var dstu2Samples = MedicationSamples.dstu2();` and then call `form()`, etc using that variable?

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationConverter.java
KevinCranmer: I believe id is correct. From the R4MedicationTransformer:
```
  List<Identifier> identifierFromProduct() {
    if (datamart.product().isEmpty()) {
      return null;
    } else {
      return Identifier.builder().id(datamart.product().get().id()).build().asList();
    }
  }
```

jhulbert-va (approved):
LGTM

jblefkove-va (approved):

PR #250, opened 2022-10-26
Remove LocationConverter.convertPosition()

KevinCranmer:
Unnecessary after converting dstu2 doubles to BigDecimals


fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/LocationConverter.java
```diff
@@ -24,7 +22,6 @@
   @Mapping(target = "status", qualifiedByName = "convertStatus")
   @Mapping(target = "mode", qualifiedByName = "convertMode")
```
jblefkove-va, line 23: Can we also replace `convertMode` method with `@Mapping(target = "type", qualifiedByName = "firstElementOrDie")` ?

PR #255, opened 2022-11-01
Api 18521 pracitioner converter

KevinCranmer:


bcastillomartel-va:
should we also add the changes to `ConverterRegistries`?

jhulbert-va (approved):

jblefkove-va (changes requested):
There is a potential `NullPointerException` for converting `active`; that field is `Boolean` in R4 and `boolean` in DSTU2

jblefkove-va (approved):

PR #257, opened 2022-11-04
Hard code category.coding.system

bcastillomartel-va:
Overwriting the R4 category system with the required DSTU2 value to pass the valid category check.


fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/ObservationConverter.java
```diff
@@ -46,6 +46,22 @@ gov.va.api.health.dstu2.api.datatypes.CodeableConcept convertAppliesTo(
         .orElse(null);
   }
 
+  @Named("convertCategory")
+  gov.va.api.health.dstu2.api.datatypes.CodeableConcept convertCategory(
+      List<gov.va.api.health.r4.api.datatypes.CodeableConcept> source) {
+    return Safe.stream(source)
+        .map(FieldConverter.INSTANCE::convert)
+        .map(
```
jblefkove-va, line 54: `forEach` is preferable to `map` for this, so that you only modify the codings directly without requiring `src.coding(...)`

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/ObservationConverter.java
```diff
@@ -46,6 +46,22 @@ gov.va.api.health.dstu2.api.datatypes.CodeableConcept convertAppliesTo(
...
+            src ->
+                src.coding(
+                    Safe.stream(src.coding())
```
jhulbert-va, line 57: The codes are all the same between r4 and dstu2?

jblefkove-va:
`convertCategory` should also include a comment to explain why the system is overridden

jblefkove-va (approved):

PR #258, opened 2022-11-07
Add Query Parameter Converting

KevinCranmer:


fhir-resource-conversion/src/test/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/PractitionerConverterTest.java
```diff
@@ -12,4 +13,13 @@ void convert() {
             json(PractitionerConverter.INSTANCE.convert(PractitionerSamples.r4().practitioner())))
         .isEqualTo(json(PractitionerSamples.dstu2().practitioner()));
   }
+
+  @Test
+  void queryParameterConverter() {
+    var r4 = QueryParameters.create().add("patient", "icn").map();
+    var dstu2 = QueryParameters.create().add("patient", "icn").map();
```
jhulbert-va, line 20: Can we make `icn` `p1` or something like that. I want to make sure there's no confusion and `icn` looks like it could be a query param itself.

KevinCranmer: Updated

jblefkove-va (approved):

PR #260, opened 2022-11-23
API-21315: FhirTokenParameter processes token parameters

jabando-va:


fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirTokenParameter.java
```diff
@@ -0,0 +1,23 @@
+package gov.va.api.health.fhir.api;
+
+import lombok.Value;
+
+@Value
+public class FhirTokenParameter {
```
jhulbert-va, line 6: I think this doesn't do enough. What if I want to build a token parameter? I think it should follow our regular format. Can be parsed, and can be created.

fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirTokenParameter.java
```diff
@@ -0,0 +1,43 @@
...
+    return FhirTokenParameter.builder().system(parts[0]).code(parts[1]).build();
+  }
+
+  public boolean systemExistsIn(String... systems) {
+    return Arrays.stream(systems).anyMatch(s -> s.equals(system()));
+  }
+
+  @Override
+  public String toString() {
+    return system() + "|" + code();
```
jblefkove-va, line 32: Should this check if `system` is null?

fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirTokenParameter.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  public static final class InvalidRequest extends IllegalArgumentException {
+    /** Invalid request exception. */
+    public InvalidRequest(String value) {
+      super(
+          String.format(
+              "invalid fhir token %s, expected format to be {system}|{code} or {code}", value));
```
jblefkove-va, line 40: Is this class intentionally excluding `{system}|` and `|{code}` ?

fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirTokenParameter.java
```diff
@@ -0,0 +1,43 @@
+package gov.va.api.health.fhir.api;
+
+import java.util.Arrays;
+import lombok.Builder;
+import lombok.Value;
+
+@Value
+@Builder
+public class FhirTokenParameter {
```
jblefkove-va, line 9: What is the distinction between this class and Vulcan [TokenParameter](https://github.com/department-of-veterans-affairs/lighthouse-vulcan/blob/main/src/main/java/gov/va/api/lighthouse/vulcan/mappings/TokenParameter.java)?

jabando-va: Ultimately they have the same roles, but we wanted Vulcan's use to be limited to applications with database-query processing like data-query/CDW. I think we decided that we did not want to use VFQ and CFQ with Vulcan, but we wanted to have some similar functionality available to them.

fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirTokenParameter.java
```diff
@@ -0,0 +1,60 @@
...
+
+  String originalString;
```
jhulbert-va, line 13: Not a fan of using string in the variable name for a string. `original` or `originalValue` would be better.

fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirTokenParameter.java
```diff
@@ -0,0 +1,60 @@
...
+
+  String system;
+
+  String code;
+
+  /** Extract system (if present) and code from parameter string. */
+  public static FhirTokenParameter parseToken(String paramString) {
+    var parts = paramString.split("\\|", -1);
+    if ((parts.length != 1 && parts.length != 2) || paramString.equals("|")) {
```
jhulbert-va, line 22: I think there should be an `isBlank` check before splitting. Then you would only need to check that the param string is not `|` and that the array length is not more than 2.

fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirTokenParameter.java
```diff
@@ -0,0 +1,60 @@
...
+      return FhirTokenParameter.builder().originalString(paramString).system(parts[0]).build();
+    }
+    return FhirTokenParameter.builder()
+        .originalString(paramString)
+        .system(parts[0])
+        .code(parts[1])
+        .build();
+  }
+
+  public boolean systemExistsIn(String... systems) {
```
jhulbert-va, line 41: `isSystemOneOf`?

fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirTokenParameter.java
```diff
@@ -0,0 +1,60 @@
...
+    return Arrays.stream(systems).anyMatch(s -> s.equals(system()));
+  }
+
+  @Override
+  public String toString() {
+    return originalString();
+  }
```
jhulbert-va, line 48:
```
FhirTokenParameter.builder().code("abc").build().toString()
```

^ That'll return null, but it should return `abc`. I'm all for storing the originalString, but we cannot assume we'll always have it.

jblefkove-va (approved):

PR #261, opened 2022-11-28
API-21315: Token Parameter processing part 2

jabando-va:


fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirTokenParameter.java
```diff
@@ -10,41 +10,53 @@
...
         .build();
   }
 
-  public boolean systemExistsIn(String... systems) {
+  public boolean isSystemOneOf(String... systems) {
     return Arrays.stream(systems).anyMatch(s -> s.equals(system()));
   }
 
   @Override
   public String toString() {
```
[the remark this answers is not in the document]
jhulbert-va: I think we may need an enum that represents where the pipe should (or shouldn't be)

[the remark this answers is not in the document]
jabando-va: parseToken populates `original`, so its toString should give back `|code`

fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirTokenParameter.java
```diff
@@ -9,42 +9,50 @@
...
-    return originalString();
+    if (isBlank(system())) {
+      return code();
+    } else if (EXPLICITLY_NONE.equals(system())) {
+      return "|" + code();
+    }
+    if (isBlank(code())) {
+      return system();
```
jhulbert-va, line 51: I think code should be `@NonNull`. If you look at the [spec for token searching](https://www.hl7.org/fhir/search.html#token) code will always either have a value or be `EXPLICITLY_NONE`. You can't have `system` only `system|` without a code.

fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirTokenParameter.java
```diff
@@ -9,42 +9,50 @@
...
-  public boolean systemExistsIn(String... systems) {
+  public boolean isSystemOneOf(String... systems) {
     return Arrays.stream(systems).anyMatch(s -> s.equals(system()));
   }
 
   @Override
   public String toString() {
-    return originalString();
+    if (isBlank(system())) {
+      return code();
```
jhulbert-va, line 46: You'll need to make sure `code` isn't `EXPLICITLY_NONE` in this case.

fhir-common/src/main/java/gov/va/api/health/fhir/api/FhirTokenParameter.java
```diff
@@ -4,47 +4,71 @@
...
+  public boolean isSystemOneOf(String... systems) {
     return Arrays.stream(systems).anyMatch(s -> s.equals(system()));
   }
 
   @Override
   public String toString() {
-    return originalString();
+    // Must have a non-blank code
+    if (isBlank(code())) {
+      throw new InvalidRequest("|");
```
jhulbert-va, line 56: I think we will want these exceptions to be better.

jabando-va: InvalidToken?

jhulbert-va (approved):

jblefkove-va (approved):

PR #264, opened 2022-12-08
API-21631-resource-conversion-mo-and-ms

jhulbert-va:
https://vajira.max.gov/browse/API-21631


fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/json/JsonNodeConverter.java
```diff
@@ -117,4 +147,16 @@ public Optional<JsonNode> suspiciousJson() {
       return Optional.empty();
     }
   }
+
+  class ShortCircuit extends Throwable {
```
bschofield-va, line 151: `Throwable` is frowned on. In general, you should extend `Exception` or `RuntimeException` (or something that extends from these). `Error` should be reserved for JVM-related problems, e.g. out of memory error. I can't think of a case where throwing (or catching) `Throwable` is desirable.

jabando-va (approved):

ilaflamme-va (approved):

jabando-va (approved):

eclendenning-va (approved):

jblefkove-va (approved):

bcastillomartel-va (approved):

PR #271, opened 2023-01-17
API-19694-spring-upgrade

jhulbert-va:
https://vajira.max.gov/browse/API-19694


argonaut-dstu2/pom.xml
```diff
@@ -25,5 +25,10 @@
       <artifactId>fhir-common</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.github.spotbugs</groupId>
+      <artifactId>spotbugs-annotations</artifactId>
```
jblefkove-va, line 30: Should we add this dependency directly to `api-starter`?

jblefkove-va (approved):

PR #272, opened 2023-01-19
Converters.firstElement grabs first element from a list

jabando-va:


fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/Converters.java
```diff
@@ -96,6 +96,15 @@ default String datetimeToDate(String datetime) {
...
   }
 
+  /** Map a single element from a list to a single object. */
+  @Named("firstElement")
+  default <T> T firstElement(Collection<T> collection) {
+    if (collection == null || collection.isEmpty()) {
+      return null;
+    }
+    return collection.stream().findFirst().orElse(null);
+  }
```
jhulbert-va, line 106: Do you have access to `Safe` here?
```
Safe.stream(collection).findFirst().orElse(null);
```

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/Converters.java
```diff
@@ -96,6 +97,12 @@ default String datetimeToDate(String datetime) {
     return DateTimeFormatter.ofPattern("yyy-MM-dd").format(fhirDateTime.atZone(ZoneId.of("UTC")));
   }
 
+  /** Map a single element from a list to a single object. */
+  @Named("firstElement")
+  default <T> T firstElement(Collection<T> collection) {
+    return Safe.stream(collection).findFirst().orElse(null);
```
jblefkove-va, line 103: update `firstElementOrDie` as well?

jabando-va: nah that one's fine I think. That one has a null check to cover the size() check that follows after.

jblefkove-va (approved):

PR #276, opened 2023-03-29
Add WellKnown duplicate properties in snake case

KevinCranmer:
https://vajira.max.gov/browse/API-25495


us-core-r4/src/main/java/gov/va/api/health/r4/api/information/WellKnown.java
```diff
@@ -31,4 +32,13 @@ public final class WellKnown {
   String introspectionEndpoint;
   String revocationEndpoint;
   @NotEmpty List<String> capabilities;
+  @NotBlank String authorization_endpoint;
```
jblefkove-va, line 35: Interleave these with the existing properties so that `authorizationEndpoint` and `authorization_endpoint` are together

Also, not thrilled with `authorization_endpoint` as a field name. How about `@JsonProperty("authorization_endpoint") String authorizationEndpoint2`

jblefkove-va:
According to http://hl7.org/fhir/smart-app-launch/1.0.0/conformance/index.html#metadata. we have fields `response_types_supported` and `token_endpoint_auth_methods`, which don't match their current camelCased versions... are either of those currently being used?

us-core-r4/src/main/java/gov/va/api/health/r4/api/information/WellKnown.java
```diff
@@ -20,15 +21,54 @@
...
   @NotBlank String tokenEndpoint;
+
+  @JsonProperty("token_endpoint")
+  @NotBlank
+  String tokenEndpoint2;
+
   List<String> tokenEndpointAuthMethodsSupported;
+
+  @JsonProperty("token_endpoint_auth_methods")
+  List<String> tokenEndpointAuthMethodsSupported2;
```
jblefkove-va, line 41: Since there is no name conflict, this can be `tokenEndpointAuthMethods` (ditto `responseTypesSupported`)

jblefkove-va (approved):

PR #281, opened 2023-03-31
API 24874 OAS update Device

bcastillomartel-va:
https://vajira.max.gov/browse/API-24874


us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Device.java
```diff
@@ -270,9 +289,16 @@ public static class DeviceName implements AsList<DeviceName>, BackboneElement {
...
-    @NotBlank String name;
+    @Schema(description = "The name of the device as given by the manufacturer.")
+    @NotBlank
+    String name;
 
-    @NotNull DeviceNameType type;
+    @Schema(
+        description =
+            "The kind or type of device. Follows "
+                + "[SNOMED code system.](http://snomed.info/sct)")
```
jblefkove-va, line 299: this should be the description for `deviceName.type`

bcastillomartel-va: fixed the mistake, thanks @jblefkove-va

jblefkove-va (approved):

PR #310, opened 2023-06-29
API-27722: MedicationOrder converter maps medref.identifier

jabando-va:


fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationOrderConverter.java
```diff
@@ -100,6 +102,25 @@ String convertDateEnded(List<gov.va.api.health.r4.api.elements.Dosage> source) {
         .orElse(null);
   }
 
+  @Named("convertMedicationReference")
+  gov.va.api.health.dstu2.api.elements.Reference convertMedicationReference(Reference source) {
+    if (source == null || source.reference() == null) {
```
jblefkove-va, line 107: Instead of verifying `source.reference()` is non-null here, would it be safer to verify that `reference` returned on line 110 is non-null?

fhir-resource-conversion/src/test/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationOrderSamples.java
```diff
@@ -61,7 +64,21 @@ public gov.va.api.health.dstu2.api.resources.MedicationOrder medicationOrder() {
                               .display("MEDICATION")
                               .build()))
                   .build())
-          .medicationReference(Reference.builder().reference("Medication/e1").build())
+          .medicationReference(
+              Reference.builder()
+                  .reference("Medication/e1")
+                  .display("Acetaminophen 325 MG Oral Tablet")
+                  .extension(
+                      List.of(
```
jblefkove-va, line 72: asList() ?

fhir-resource-conversion/src/main/java/gov/va/api/health/fhir/conversion/r4todstu2/resources/MedicationOrderConverter.java
jabando-va: Not sure about "safer". I think the current way makes sense, null checks to return null if needed: the "return early" approach. I cant see a reason to attempt converting the field before the null checks.

jblefkove-va (approved):

